PR #4, opened 2019-06-17
Add the _FieldSet type, make it the type for the input argument for @requires, @provides and @key

mcohen75:
Here I've introduced the _FieldSet type and made it the type of the input argument for the @requires, @provides and @key directives. Addresses #5.

When adding the scalar type for _FieldSet to the type wiring, the _FieldSet scalar type is present in the original schema and is printed when generating the `sdl` field for the `_service` query (i.e. fetch service capabilities).

graphql-java's `SchemaPrinter` does not have the ability to exclude certain types from being printed. Since the schema returned by the _service query must exclude federation specific types and directives, we must find a way to exclude this type.

To do so I've taken the following approach:
1. Create the original schema without federation types. Note that in the `SchemaUtils` test helper I've removed the code to add definitions and moved it to the SchemaTransformer.
2. Create the original schema with the option to *not* enforce that directives are defined/correct.

If we want to enforce directives on initial schema creation, we can:
1. Implement our own SchemaPrinter :(
2. Lobby for modifications to SchemaPrinter to allow filtering of types or some other way to further customize the output.

My thought going forward is that we can create an abstraction for creating federated schemas that accepts `TypeDefinitionRegistry` and `TypeWiring` instances.

The combined approach has some advantages:
1. It simplifies federated schema creation for consumers. Consumers would just use `FederatedSchemaGenerator` instead of `SchemaGenerator`.
2. It is Spring Boot friendly. A starter can be created that consumes the `TypeDefinitionRegistry` and `TypeWiring` instances and creates a federated schema.
3. Schema creation logic is more encapsulated and can be changed/fixed. For example, if we would like to change the code to enforce the presence of directive definitions when creating schema, we can change this once we implement one of the alternative paths above.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -15,7 +15,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
-import static graphql.Scalars.GraphQLString;
 import static graphql.introspection.Introspection.DirectiveLocation.FIELD_DEFINITION;
 import static graphql.introspection.Introspection.DirectiveLocation.INTERFACE;
 import static graphql.introspection.Introspection.DirectiveLocation.OBJECT;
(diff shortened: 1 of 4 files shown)


mcohen75:
cc: @pcarrier @zionts

pcarrier:
This change looks good to me. Discussing `TypeDefinitionRegistry`+`TypeWiring` vs `GraphQLSchema` in Spectrum; curious to provide a better option without affecting non-IDL developers.

pcarrier:
Would love to merge this; could you push to that branch to trigger our PR hooks? They weren't correctly set up when you opened the PR. Also reached out on Spectrum.

mcohen75:
@pcarrier I have signed the CLA.

Issue #7, opened 2019-06-20
Schema transformation fails when two schemas that have a type with the same name are transformed


mcohen75:
When two schemas are transformed that have a type with the same name, schema transformation fails with an error indicating that duplicate types are not allowed.

```
graphql.AssertException: All types within a GraphQL schema must have unique names. No two provided types may have the same name.
No provided type may have a name which conflicts with any built in types (including Scalar and Introspection types).
You have redefined the type 'SomeType' from being a 'GraphQLObjectType' to a 'GraphQLObjectType'
```

This appears to be caused by the fact that there is just one instance of the _Entity type definition. I'm not sure of the precise mechanism, but somewhere along the way `graphql-java` appears to be modifying types and/or definition instances in such a way that retains references from prior schemas.

To avoid this I think it may be best to always create new type and definition instances when transforming schemas.

With a schema like:
```graphql
type Query {
}

type SomeType @key(fields: "theKey") {
  theKey: ID!
  field1: String!
  field2: Int!
}
```

The following test reproduces the problem:
```java
@Test
void testSchemaTransformationIsolated() {

    final GraphQLSchema schema = SchemaUtils.buildSchema(TestUtils.readResource("isolated.graphql"));

    final GraphQLSchema federatedSchema1 = Federation.transform(schema)
            .resolveEntityType(env -> null)
            .fetchEntities(environment -> null)
            .build();

    final GraphQLSchema schema2 = SchemaUtils.buildSchema(TestUtils.readResource("isolated.graphql"));

    final GraphQLSchema federatedSchema2 = Federation.transform(schema2)
            .resolveEntityType(env -> null)
            .fetchEntities(environment -> null)
            .build();
}
```

The failure:
```
graphql.AssertException: All types within a GraphQL schema must have unique names. No two provided types may have the same name.
No provided type may have a name which conflicts with any built in types (including Scalar and Introspection types).
You have redefined the type 'SomeType' from being a 'GraphQLObjectType' to a 'GraphQLObjectType'

	at graphql.schema.GraphQLTypeCollectingVisitor.assertTypeUniqueness(GraphQLTypeCollectingVisitor.java:105)
	at graphql.schema.GraphQLTypeCollectingVisitor.visitGraphQLObjectType(GraphQLTypeCollectingVisitor.java:38)
	at graphql.schema.GraphQLObjectType.accept(GraphQLObjectType.java:163)
[...]

PR #8, opened 2019-06-24
v0.1.0: valued directive creation, spec update, bugfix, CI fix

pcarrier:
- Make some names public, when likely to be accessed
- Make it easy to construct directives with values
- `@external` isn't allowed on `OBJECT`
- Fix a bug around building multiple schemas (`_Entity.field` was mutated during schema builds to resolve type references)
- Fix `pom.xml` files for bintray uploads
- Drop release plugin configuration

Please see individual commits for details.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -26,82 +26,136 @@ import static graphql.schema.GraphQLDirective.newDirective;
 
 @PublicApi
 public final class FederationDirectives {
+    /* Directive locations */
+
+    private static final DirectiveLocation DL_OBJECT = newDirectiveLocation()
+            .name("OBJECT")
+            .build();
+
+    private static final DirectiveLocation DL_INTERFACE = newDirectiveLocation()
+            .name("INTERFACE")
+            .build();
+
+    private static final DirectiveLocation DL_FIELD_DEFINITION = newDirectiveLocation()
+            .name("FIELD_DEFINITION")
+            .build();
+
+    /* fields: _FieldSet */
+
     private static final GraphQLArgument fieldsArgument = newArgument()
             .name("fields")
             .type(new GraphQLNonNull(_FieldSet.type))
             .build();
+
+    private static final GraphQLArgument fieldsArgument(String value) {
+        return newArgument(fieldsArgument)
+                .value(value)
+                .build();
+    }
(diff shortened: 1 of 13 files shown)


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -26,82 +26,136 @@
...
+
     public static final String requiresName = "requires";
+
     public static final GraphQLDirective requires = newDirective()
             .name(requiresName)
             .validLocations(FIELD_DEFINITION)
             .argument(fieldsArgument)
             .build();
+
+    public static final GraphQLDirective requires(String fields) {
```
zionts: usually public APIs have some kind of javadoc, right? We probably don't need it to explain the spec, but we could add a note to usage and maybe link to the spec doc?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/_Any.java
```diff
@@ -21,8 +21,8 @@
 
 import static graphql.schema.GraphQLScalarType.newScalar;
 
-final class _Any {
-    static final String typeName = "_Any";
+public final class _Any {
```
zionts: out of curiosity, what's the use case for this being public?

pcarrier: It seems worth making it reusable outside of the library.

mcohen75:
It looks like some of the changes here are meant to address #7. Would it be worth adding a test for this? It's pretty straight forward as you just need to create more than one schema.

pcarrier:
@mcohen75 Indeed! Copy-pasted your test.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/_Any.java
mcohen75: It might be necessary in order to provide support for projects that don't use SDL to create their schema. As @pcarrier pointed out [GraphQL Kotlin](https://github.com/ExpediaDotCom/graphql-kotlin) is one such project that seems to be gaining traction.

That said, it does add significant liability. These can't be changed or moved easily if they remain public. It might be good to look at other mechanisms for sharing the functionality here, like [Hooks in GraphQL Kotlin](https://github.com/ExpediaDotCom/graphql-kotlin/wiki/Schema-Generator-Hooks).

PR #10, opened 2019-07-02
Testing: Update settings.xml

abernix:
This is not meant to be merged, merely to confirm that CircleCI jobs are running correctly and that status checks are being displayed on the PR correctly.


abernix:
Seems like it worked...

image

PR #11, opened 2019-07-09
Federation.transform shorthands for SDL users, optional query type, custom _Any type, concrete types in _Entity

pcarrier:
Please see individual commits.

The new `Federation.transform` signatures should simplify usage for SDL users by quite a bit, as illustrated in tests and the Spring demo.

Schemas coming in do not need to have a query type, we can create one as needed.

Schemas might want to use `_Any` directly themselves, or specify a coercing for the entities type resolver and data fetcher.


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
```diff
@@ -1,14 +1,48 @@
...
 
     @NotNull
     public static SchemaTransformer transform(final GraphQLSchema schema) {
-        return new SchemaTransformer(schema);
+        return new SchemaTransformer(() -> schema);
+    }
+
+    public static SchemaTransformer transform(final TypeDefinitionRegistry typeRegistry, final RuntimeWiring runtimeWiring) {
+        return transform(new SchemaGenerator().makeExecutableSchema(
+                SchemaGenerator.Options.defaultOptions()
```
zionts: why not use the `generatorOptions` from above here?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -28,136 +29,89 @@
...
+            .name(extendsName)
+            .validLocations(OBJECT)
+            .build();
+    public static final Set<GraphQLDirective> allDirectives = new HashSet<>();
+    public static final Set<SDLDefinition> allDefinitions = new HashSet<>();
     private static final DirectiveLocation DL_OBJECT = newDirectiveLocation()
             .name("OBJECT")
             .build();
 
+    /* directive @external on FIELD_DEFINITION */
```
zionts: Something is wonky here? This comment is referring to the `@external` directive, but the below line is on the `@extends` directive.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -28,136 +29,89 @@
...
     private static final DirectiveLocation DL_FIELD_DEFINITION = newDirectiveLocation()
             .name("FIELD_DEFINITION")
             .build();
 
-    /* fields: _FieldSet */
-
-    private static final GraphQLArgument fieldsArgument = newArgument()
-            .name("fields")
-            .type(new GraphQLNonNull(_FieldSet.type))
+    /* directive @requires(fields: _FieldSet!) on FIELD_DEFINITION */
```
zionts: Similarly, this comment seems mismatched.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -28,136 +29,89 @@
...
-        return newArgument(fieldsArgument)
-                .value(value)
-                .build();
-    }
-
-    private static final InputValueDefinition fieldsDefinition = newInputValueDefinition()
+    private static final GraphQLArgument fieldsArgument = newArgument()
             .name("fields")
-            .type(new NonNullType(new TypeName(_FieldSet.typeName)))
+            .type(GraphQLNonNull.nonNull(new GraphQLTypeReference(_FieldSet.typeName)))
```
zionts: I know this can sometimes have some strange behavior in relying on reflection, and though I think this is fine, I'm always a little cautious in relying on it if we don't have to, and I'm not sure why we need this to be a type reference rather than the type itself?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -28,136 +29,89 @@
...
-        return newDirective(provides)
-                .argument(fieldsArgument(fields))
-                .build();
-    }
-
-    public static final DirectiveDefinition providesDefinition = newDirectiveDefinition()
-            .name(providesName)
+    private static final InputValueDefinition fieldsDefinition = newInputValueDefinition()
+            .name("fields")
+            .type(new NonNullType(new TypeName(_FieldSet.typeName)))
```
zionts: Seems like we might be able to reuse this type and be more consistent over TypeReference, TypeName, or the actual type.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -28,136 +29,89 @@
...
+            .directiveLocations(Arrays.asList(DL_OBJECT, DL_INTERFACE))
+            .inputValueDefinition(fieldsDefinition)
+            .build();
+    public static final DirectiveDefinition requiresDefinition = newDirectiveDefinition()
+            .name(requiresName)
             .directiveLocations(Arrays.asList(DL_FIELD_DEFINITION))
             .inputValueDefinition(fieldsDefinition)
             .build();
 
     /* directive @extends on OBJECT */
```
zionts: comment is also incorrect here

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -53,14 +54,21 @@ public SchemaTransformer fetchEntitiesFactory(DataFetcherFactory entitiesDataFet
     @NotNull
     public final GraphQLSchema build() throws SchemaProblem {
         final List<GraphQLError> errors = new ArrayList<>();
-        final GraphQLSchema.Builder schema = GraphQLSchema.newSchema(originalSchema);
+        final GraphQLSchema original = schemaSupplier.get();
 
-        final GraphQLObjectType originalQueryType = originalSchema.getQueryType();
+        final GraphQLSchema.Builder schema = GraphQLSchema.newSchema(original);
```
zionts: nit: rename to schemaCopy?

spring-example/src/test/java/com/apollographql/federation/springexample/AppTest.java
```diff
@@ -0,0 +1,31 @@
...
+@RunWith(SpringRunner.class)
+@GraphQLTest
+public class AppTest {
+    @Autowired
+    private GraphQLTestTemplate graphqlTestTemplate;
+
+    @Test
+    public void lookupPlanckProduct() throws IOException {
+        final GraphQLResponse response = graphqlTestTemplate.postForResource("LookupPlanckProduct.graphql");
+        assertNotNull(response, "response not null");
```
zionts: nit: This might end up in a double negative if it fails. Might be better to use the positive `response is defined` here.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
pcarrier: Funnily enough because `graphql.language.InputValueDefinition.Builder#type` wants nothing to do with the `graphql.schema` concept of a type.

Those 2 concepts are only connected through `Object`:

Type

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
pcarrier: Oops, code got "reordered" :(

spring-example/src/test/java/com/apollographql/federation/springexample/AppTest.java
pcarrier: What do you think of the new messages?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
pcarrier: Reuse it where?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
zionts: sad times.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
zionts: Seems like it's gone

zionts:
This looks great! Loving the additions for interface entity and the query type allowance

PR #25, opened 2019-08-29
feat: add federated tracing instrumentation

### LINKED ISSUE #24: Federated tracing support
Is anyone working on support for [extensions.ftv1](https://www.apollographql.com/docs/apollo-server/federation/metrics/) tracing yet?

lennyburdette:
Federated GraphQL services should include timing and error
information as a Base64-encoded protocol buffer message in
the "extensions.ftv1" field.

This change includes a tracer that uses the graphql-java
instrumentation API to record field timings and info and attach it
on the execution context.

I used the Apollo Server typescript code as reference:

https://github.com/apollographql/apollo-server/blob/master/packages/apollo-engine-reporting/src/federatedExtension.ts
https://github.com/apollographql/apollo-server/blob/master/packages/apollo-engine-reporting/src/treeBuilder.ts

As well as my ruby implementation:

https://github.com/lennyburdette/apollo-federation-ruby/blob/federated-tracing/lib/apollo-federation/tracing/tracer.rb

Federated tracing documentation: https://www.apollographql.com/docs/apollo-server/federation/metrics/

Fixes #24

### DIFF
--- a/graphql-java-support/pom.xml
+++ b/graphql-java-support/pom.xml
@@ -30,9 +30,21 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.9.0</version>
+        </dependency>
     </dependencies>
 
     <build>
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>1.6.0</version>
+            </extension>
+        </extensions>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
@@ -53,6 +65,23 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>cobertura-maven-plugin</artifactId>
+            </plugin>
+
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>0.6.1</version>
+                <configuration>
+                    <protocArtifact>com.google.protobuf:protoc:3.9.0:exe:${os.detected.classifier}</protocArtifact>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>compile</goal>
+                            <goal>test-compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
@@ -11,6 +11,7 @@ import graphql.execution.ExecutionStepInfo;
 import graphql.execution.instrumentation.InstrumentationContext;
 import graphql.execution.instrumentation.InstrumentationState;
 import graphql.execution.instrumentation.SimpleInstrumentation;
+import graphql.execution.instrumentation.parameters.InstrumentationCreateStateParameters;
 import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters;
 import graphql.execution.instrumentation.parameters.InstrumentationFieldFetchParameters;
 import graphql.execution.instrumentation.parameters.InstrumentationValidationParameters;
@@ -20,6 +21,7 @@ import graphql.parser.InvalidSyntaxException;
 import graphql.validation.ValidationError;
 import mdg.engine.proto.Reports;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,6 +38,9 @@ import static graphql.schema.GraphQLTypeUtil.simplePrint;
 
 public class FederatedTracingInstrumentation extends SimpleInstrumentation {
     private static final String EXTENSION_KEY = "ftv1";
+    private static final String HEADER_NAME = "apollo-federation-include-trace";
+    private static final String HEADER_VALUE = "ftv1";
+
     private final Options options;
 
     private static final Logger logger = LoggerFactory.getLogger(FederatedTracingInstrumentation.class);
@@ -49,13 +54,30 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
     }
 
     @Override
-    public InstrumentationState createState() {
+    public InstrumentationState createState(InstrumentationCreateStateParameters parameters) {
+        // If we've been configured with a way of reading HTTP headers, we should only be active
+        // if the special HTTP header has the special value. If the header isn't provided or has
+        // a different value, return null for our state, which we'll interpret in the rest of this
+        // file as meaning "don't instrument".  (If we haven't been given access to HTTP headers,
+        // always instrument.)
+        Object context = parameters.getExecutionInput().getContext();
+        if (context instanceof HTTPRequestHeaders) {
+            @Nullable String headerValue = ((HTTPRequestHeaders) context).getHTTPRequestHeader(HEADER_NAME);
+            if (headerValue == null || !headerValue.equals(HEADER_VALUE)) {
+                return null;
+            }
+        }
         return new FederatedTracingState();
     }
 
     @Override
     public CompletableFuture<ExecutionResult> instrumentExecutionResult(ExecutionResult executionResult, InstrumentationExecutionParameters parameters) {
-        Reports.Trace trace = parameters.<FederatedTracingState>getInstrumentationState().toProto();
+        final @Nullable FederatedTracingState state = parameters.<FederatedTracingState>getInstrumentationState();
+        if (state == null) {
+            return super.instrumentExecutionResult(executionResult, parameters);
+        }
+
+        Reports.Trace trace = state.toProto();
 
         if (options.isDebuggingEnabled()) {
             logger.debug(trace.toString());
@@ -70,6 +92,10 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
     @Override
     public InstrumentationContext<Object> beginFieldFetch(InstrumentationFieldFetchParameters parameters) {
         FederatedTracingState state = parameters.getInstrumentationState();
+        if (state == null) {
+            return super.beginFieldFetch(parameters);
+        }
+
         SourceLocation fieldLocation = parameters.getEnvironment().getField().getSourceLocation();
 
         long startNanos = System.nanoTime();
@@ -92,6 +118,10 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
     @Override
     public InstrumentationContext<Document> beginParse(InstrumentationExecutionParameters parameters) {
         FederatedTracingState state = parameters.getInstrumentationState();
+        if (state == null) {
+            return super.beginParse(parameters);
+        }
+
         return whenCompleted((document, throwable) -> {
             for (GraphQLError error : convertErrors(throwable, null)) {
                 state.addRootError(error);
@@ -102,6 +132,10 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
     @Override
     public InstrumentationContext<List<ValidationError>> beginValidation(InstrumentationValidationParameters parameters) {
         FederatedTracingState state = parameters.getInstrumentationState();
+        if (state == null) {
+            return super.beginValidation(parameters);
+        }
+
         return whenCompleted((validationErrors, throwable) -> {
             for (GraphQLError error : convertErrors(throwable, null)) {
                 state.addRootError(error);
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
@@ -86,6 +86,7 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
 
     // Field resolvers can throw exceptions or add errors to the DataFetchingResult. This method normalizes them to a
     // single list of GraphQLErrors.
+    @NotNull
     private List<GraphQLError> convertErrors(Throwable throwable, Object result) {
         ArrayList<GraphQLError> graphQLErrors = new ArrayList<>();
 
@@ -126,6 +127,7 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
             nodesByPath.put(ExecutionPath.rootPath(), Reports.Trace.Node.newBuilder());
         }
 
+        @NotNull
         Reports.Trace toProto() {
             return Reports.Trace.newBuilder()
                     .setStartTime(getStartTimestamp())
@@ -172,6 +174,7 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
             nodesByPath.put(path, node);
         }
 
+        @NotNull
         Reports.Trace.Node.Builder getParent(ExecutionPath path) {
             List<Object> pathParts = path.toList();
             ExecutionPath parentPath = ExecutionPath.fromList(pathParts.subList(0, pathParts.size() - 1));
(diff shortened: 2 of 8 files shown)


lennyburdette:
After a _really fun_ day of learning maven and fighting with a broken CI system, I have evidence that this change with Apollo tracing!

Screen Shot 2019-08-30 at 1 51 39 PM

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+            return System.nanoTime() - startRequestNanos;
+        }
+
+        private Reports.Trace.Node buildNodeTree() {
+            return nodeMap.get(ExecutionPath.rootPath()).buildRecursive();
+        }
+    }
+
+    /**
+     * Wrapper for Node information with mutable lists for children and errors.
```
zionts: Do we need to include this? I believe that the builder has accessors for adding children and errors.

graphql-java-support/src/main/proto/reports.proto
```diff
@@ -0,0 +1,488 @@
+syntax = "proto3";
```
zionts: I think it would be useful to include a comment along the lines of 
```md
This file is copied over from the `apollo-engine-reporting` package, which, in turn, derives its information from Apollo's cloud infrastructure. It is expected that as the protobuf definition changes within the underlying system we are reporting to, these libraries will need to be updated in sync.
```

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+    public InstrumentationState createState() {
+        return new FederatedTracingState();
+    }
+
+    @Override
+    public CompletableFuture<ExecutionResult> instrumentExecutionResult(ExecutionResult executionResult, InstrumentationExecutionParameters parameters) {
+        FederatedTracingState state = parameters.getInstrumentationState();
+
+        Map<Object, Object> extensions = executionResult.getExtensions();
+        Map<Object, Object> tracingMap = new LinkedHashMap<>(extensions == null ? Collections.emptyMap() : extensions);
```
zionts: This is a confusing name for something that is actually a copy of the extensions that are already present within the execution result and has nothing to do with tracing until we add in our own extension below. Maybe something like `extensionsCopy` or `currentExtensionsMap` instead?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+
+        Reports.Trace trace = state.toProto();
+
+        byte[] bytes = trace.toByteArray();
```
zionts: nbd: can we inline this?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+        tracingMap.put(KEY, Base64.getEncoder().encodeToString(bytes));
+
+        if (debuggingEnabled) {
```
zionts: I like this feature, though I think it might be more useful as a log than as a separate extensions key 🤷‍♂ curious on your thoughts!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+        return whenCompleted((result, throwable) -> {
+            long endNanos = System.nanoTime();
+
+            DataFetchingEnvironment env = parameters.getEnvironment();
+            ExecutionStepInfo executionStepInfo = env.getExecutionStepInfo();
+
+            Node node = state.addNode(executionStepInfo, startNanos, endNanos);
+
+            if (throwable != null) {
+                node.addError(throwable);
```
zionts: feat: It might be useful to allow implementors to decide how they want to translate thrown errors into GraphQL errors that services report to the gateway, but we can definitely do that when we actually have a feature request and not during this round of code review 😉

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+import static graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted;
+import static graphql.schema.GraphQLTypeUtil.simplePrint;
+
+public class FederatedTracingInstrumentation extends SimpleInstrumentation {
+    private static final String KEY = "ftv1";
```
zionts: nbd: `EXTENSION_KEY`?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+        private final long startRequestNanos;
+        private final LinkedHashMap<ExecutionPath, Node> nodeMap;
+
+        private FederatedTracingState() {
+            // record start time when creating instrumentation state for a request
+            startRequestTime = Instant.now();
+            startRequestNanos = System.nanoTime();
+
+            // Create node map with the root node
+            nodeMap = new LinkedHashMap<>();
```
zionts: naming suggestion: `nodesByPath`?

zionts: Also it seems like this state could likely be largely simplified to be built using the tree from `TraceNode.newBuilder`. I think that this object also means that building the node tree is a more arduous task at the end, and it would likely be a quick performance win (and code cleanliness win) to build the object directly rather than through a wrapper.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+            return Reports.Trace.newBuilder()
+                    .setStartTime(getStartTimestamp())
+                    .setEndTime(getNowTimestamp())
+                    .setDurationNs(getDuration())
+                    .setRoot(buildNodeTree())
+                    .build();
+        }
+
+        /**
+         * Adds node to nodeMap and recursively ensures that all parent node exist.
```
zionts: typo: nodes

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+         */
+        Node addNode(ExecutionPath path, Node node) {
+            if (!nodeMap.containsKey(path)) {
+                nodeMap.put(path, node);
+            }
+
+            List<Object> pathParts = path.toList();
+            ExecutionPath parentPath = ExecutionPath.fromList(pathParts.subList(0, pathParts.size() - 1));
+
+            if (!nodeMap.containsKey(parentPath)) {
```
zionts: Out of curiosity, since I'm only giving this a brief look atm: how can we get into a state like this? Would this be caused from a bug within the instrumentation pipeline around field execution in GraphQL Java?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,282 @@
...
+                addNode(parentPath, new Node(parentPath));
+            }
+
+            nodeMap.get(parentPath).addChild(node);
```
zionts: oh hm. I didn't realize we were indexing the fields by path **and** creating their children. I suppose that's relatively low memory overhead because we're just dealing with pointers underneath, but I haven't gotten to the point yet that necessitates indexing by field or justifies the value in it.

lennyburdette: I'm using the same strategy that I found in [this fork](https://github.com/salsify/apollo-tracing-ruby/blob/feature/new-apollo-api/lib/apollo_tracing/tracer.rb) that adds non-federated apollo tracing to the ruby gem. Pretty much just rewrote it in Java here!

You're probably right though, and removing the additional object created for each node would be a performance win. I'll take a look and come back to you if I can't figure out a way to do it!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
lennyburdette: Yeah, I left that feature out of the ruby implementation too. One thing I might do today is make it easier to add configuration options for tracing, which would give us a place to add a error translation hook.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
lennyburdette: The reason for the backwards construction of node parents is to support indexes. The instrumentation hooks don't fire for `widgets.0`, just for `widgets.0.id`. I agree this isn't super clear though. I'll see what happens to this part of the code when I try dropping the wrapper class.

lennyburdette:
@zionts You were right, no need for a wrapper class at all!

The secret sauce was discovering the `builder.addChildBuilder()` API, which allows me to add children and errors to each node without building them. I didn't know that protos would build recursively ... very cool!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,228 @@
...
+                    .setNanos(now.getNano())
+                    .build();
+        }
+
+        private long getDuration() {
+            return System.nanoTime() - startRequestNanos;
+        }
+    }
+
+    public static class Options {
```
zionts: nice

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        Reports.Trace.Node.Builder getParent(ExecutionPath path) {
+            List<Object> pathParts = path.toList();
+            ExecutionPath parentPath = ExecutionPath.fromList(pathParts.subList(0, pathParts.size() - 1));
+
+            if (!nodesByPath.containsKey(parentPath)) {
+                Reports.Trace.Node.Builder missingParent = getParent(parentPath).addChildBuilder();
+
+                // Missing parents are always list items, so we need to add the `index` field to them
```
zionts: Why is this true? Is this relying on the way that GraphQL-Java executes? And why isn't it true for list items?

glasser: I assume it's based on the assumption that execution goes from top to bottom.

Note though that GraphQL types can be nested lists – is that what you're asking about in the second sentence?

zionts: that makes sense, and I think that's true, but I'm questioning whether that's a safe assumption to bake in and if not, how we might avoid it

lennyburdette: There is an instrumentation hook called [beginFieldListComplete](https://github.com/graphql-java/graphql-java/blob/b5781e3bbaeababe4bb916cf6acba6763fd7c875/src/main/java/graphql/execution/instrumentation/Instrumentation.java#L154). It seemed promising because there's a `getCurrentListIndex()` as part of the arguments to the hook. But it doesn't work: it's called once per list, not once per list item. ([Calling code here](https://github.com/graphql-java/graphql-java/blob/735ce0626e608ad1c6530924f8076851aa18f70b/src/main/java/graphql/execution/ExecutionStrategy.java#L500)).

I think this is the best bet. I didn't consider nested lists, so I added a test for them and it seems to work fine!

If y'all can figure out another way to create the indexed nodes, I'm all ears. Maybe I'll post something on the graphql-java spectrum to see if one of the maintainers have any ideas.

lennyburdette: https://spectrum.chat/graphql-java/general/instrumentation-hook-for-list-items~534fa292-73b0-4cce-8832-5bbda66dcfb3

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -0,0 +1,255 @@
...
+
+            String originalFieldName = stepInfo.getField().getName();
+
+            // set originalFieldName only when a field alias was used
+            if (!originalFieldName.equals(stepInfo.getResultKey())) {
+                node.setOriginalFieldName(originalFieldName);
+            }
+
+            errors.forEach(error -> {
+                Reports.Trace.Error.Builder builder = node.addErrorBuilder().setMessage(error.getMessage());
```
glasser: The JS implementation also takes other kinds of errors and puts them on the root node — most notably parse and validation errors, which the instrumentation can instrument. Do you want to implement that too?

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/FederatedTracingInstrumentationTest.java
```diff
@@ -0,0 +1,155 @@
...
+        TypeDefinitionRegistry typeDefs = new SchemaParser().parse(tracingSDL);
+        RuntimeWiring resolvers = RuntimeWiring.newRuntimeWiring()
+                .type("Query", builder ->
+                        // return two items
+                        builder.dataFetcher("widgets", env -> {
+                            ArrayList<Object> objects = new ArrayList<>(2);
+                            objects.add(new Object());
+                            objects.add(new Object());
+                            return objects;
+                        }).dataFetcher("listOfLists", env -> {
```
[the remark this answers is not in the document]
lennyburdette: Whoops, forgot to check in that file. Thanks for catching!

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/FederatedTracingInstrumentationTest.java
```diff
@@ -0,0 +1,155 @@
...
+                            lists.get(1).add(new Object());
+                            lists.get(1).add(new Object());
+                            return lists;
+                        })
+                                .dataFetcher("listOfScalars", env -> new String[]{"one", "two", "three"}))
+                .type("Widget", builder ->
+                        // Widget.foo works normally, Widget.bar always throws an error
+                        builder.dataFetcher("foo", env -> "hello world")
+                                .dataFetcher("bar", env -> {
+                                    ExceptionWhileDataFetching whoops = new ExceptionWhileDataFetching(
```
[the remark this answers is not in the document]
lennyburdette: :+1: passing in the field location from the instrumentation around the resolver is way better than depending on a specific kind of error!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
lennyburdette: Great idea, this was pretty easy. I'm not entirely sure if I'm grabbing locations off the throwable in the best way (in `convertErrors`) ... got any other ideas?

glasser: Looks good enough to me

glasser:
Would you like to do the README and http headers change or should I?

Here's what I think would make a good interface for the header thing:

```
interface HTTPRequestHeaders {
  @Nullable String getHTTPRequestHeader(String caseInsensitiveHeaderName);
}
```

lennyburdette:
@glasser feel free to jump on it! I’m focused on pitching an investment in Apollo to the higher-ups and probably won’t get back to this for a week. 😄

graphql-java-support/pom.xml
```diff
@@ -62,6 +74,23 @@
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>cobertura-maven-plugin</artifactId>
             </plugin>
+
+            <plugin>
```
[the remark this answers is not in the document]
pcarrier: Neithr am I, but this seems to match https://www.xolstice.org/protobuf-maven-plugin/usage.html so LGTM.

glasser:
Enabled CI testing of forked PRs, which would have caught some of the earlier issues!

PR #26, opened 2019-09-13
spring-example pom.xml: assembly plugin

pcarrier:
Allows for:

```
$ mvn package && java -jar spring-example/target/federation-spring-example-0.2.1-SNAPSHOT-jar-with-dependencies.jar
```

We should confirm the artifacts do not get attached for upload.

Should update README.md to reflect the option.

### DIFF
--- a/spring-example/pom.xml
+++ b/spring-example/pom.xml
@@ -75,6 +75,7 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


lqc:
I see you already list "springboot-maven-plugin", so why not use that (https://docs.spring.io/spring-boot/docs/current/maven-plugin/repackage-mojo.html) instead. The repackage mojo is the way recommended by Spring for this.

pcarrier:
@lqc Sounds better, thanks, dropping this.

PR #32, opened 2019-10-17
Fix Compatibility With graphql-java-tools

mkaulig:
Allow transformation of schemas which have the query type both
explicitly as queryType and also in additional types. If not both types
are replaced consistently by the new query type the validation of the
resulting schema fails

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
@@ -98,12 +98,18 @@ public final class SchemaTransformer {
                 .map(GraphQLType::getName)
                 .collect(Collectors.toSet());
 
+        Set<GraphQLType> newAdditionalTypes = originalSchema.getAdditionalTypes()
+                .stream()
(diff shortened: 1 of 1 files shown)


havenotfear:
Any update on this?

pcarrier:
@havenotfear Thanks for the contribution! Could you please write a test for the problem you're trying to solve, and/or explain what it was?

mkaulig:
@pcarrier I already made a PR in the graphql-java-tools project to fix the underlying issue in graphql-java-tools here https://github.com/graphql-java-kickstart/graphql-java-tools/pull/327 . With the next version of graphql-java-tools this fix should be released and the resulting issue in federation-jvm is also fixed with this new version. I think my Pull Request here can be closed.

The problem I was trying to solve is the following:

graphql-java-tools generates a schema where the query type exists in both the queryType and also in the additionalTypes of the GraphQLSchema instance. This is accepted by graphql-java as a valid schema since the multiple query types are the same instance.
When transforming the query type for federation, only the one query type in the queryType field of the GraphQLSchema gets transformed. This leads to two different query types in the Schema (one in queryType and a different one in additionalTypes) which is not accepted by graphql-java as valid.

PR #33, opened 2019-10-25
Change for some added examples similar js and graphql-java-tools compatibilty

amarflybot:
used https://github.com/apollographql/federation-demo to build an equivalent in java.


.idea/sonarlint/issuestore/9/6/96a170b5ff9c9d52dbad6ded97c0a8feaee83d70
```diff
@@ -0,0 +1,17 @@
+
+u2squid:RightCurlyBraceDifferentLineAsNextBlockCheck�"1Move this "else" keyword to a new dedicated line.(�������
```
zionts: Is there a good reason to check this in?

.gitignore
```diff
@@ -1,3 +1,71 @@
-/.idea/workspace.xml
-/.mvn/wrapper/maven-wrapper.jar
+
```
pcarrier: Could you please take out unrelated plumbing changes?

zionts:
I like the idea of making the different spring examples build-able into their own jars 👍 I agree with Pierre that we should remove the changes to `.idea` files and to the `.gitignore` in this PR, though. If there's a good reason for them, we would welcome another PR!

spring-example-product/src/main/java/com/apollographql/federation/product/App.java
```diff
@@ -0,0 +1,19 @@
+package com.apollographql.federation.product;
```
pcarrier: Could you please make sure that every subproject (eg `spring-example-product`) has its own package (eg `com.apollographql.federation.springexample.product`, no strong opinion on last level of nesting)?

spring-example-user/src/main/resources/application.yml
```diff
@@ -0,0 +1,17 @@
+spring:
+  application:
+    name: federation-jvm-spring-example
```
pcarrier: Shouldn't each app have its own name? Otherwise, we should be able to share resources like this across examples using a common module.

pcarrier:
Many thanks! Could you clarify how graphql-java-tools compatibility is addressed in the PR description?

spring-example-user/src/main/resources/application.yml
amarflybot18: Yes this is a miss, all apps should have individual Spring application names.

hwillson:
👋 @amarflybot - are you still interested in getting these changes in? If you're able to rebase and address the outstanding review comments, we'll get this merged. Thanks!

hwillson:
@amarflybot actually, talking things over with the team it looks like most of this work has already been completed in one form or another, so I'll close this off. Thanks for your time on this!

PR #40, opened 2020-01-17
Refactored FederatedTracingState to be thread-safe

sachindshinde:
This PR is to make the `FederatedTracingState` class be thread-safe, due to issues mentioned in https://github.com/apollographql/federation-jvm/pull/38#issuecomment-575330134 .

This PR basically refactors the recursion into a loop, so we can reduce the code in `computeIfAbsent`.


sachindshinde:
Closed this PR because I realized it doesn't really go far enough in ensuring thread-safety. The revision to this PR is #42 .

PR #41, opened 2020-01-19
Bump graphql-java to v14

williamboman:


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -85,7 +95,7 @@ public final GraphQLSchema build() throws SchemaProblem {
         final Set<String> entityTypeNames = originalSchema.getAllTypesAsList().stream()
                 .filter(t -> t instanceof GraphQLDirectiveContainer &&
                         ((GraphQLDirectiveContainer) t).getDirective(FederationDirectives.keyName) != null)
-                .map(GraphQLType::getName)
+                .map(GraphQLNamedType::getName)
```
williamboman: https://github.com/graphql-java/graphql-java/pull/1605

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -138,8 +148,9 @@ private String sdl() {
         final SchemaPrinter.Options options = SchemaPrinter.Options.defaultOptions()
                 .includeScalarTypes(true)
                 .includeExtendedScalarTypes(true)
-                .includeSchemaDefintion(true)
-                .includeDirectives(true);
+                .includeSchemaDefinition(true)
+                .includeDirectives(directivePredicate);
```
williamboman: https://github.com/graphql-java/graphql-java/pull/1721

graphql-java-support/src/test/resources/com/apollographql/federation/graphqljava/schemas/directives-output.graphql
```diff
@@ -0,0 +1,32 @@
...
+directive @skip(
+    "Skipped when true."
+    if: Boolean!
+  ) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+"Marks the field or enum value as deprecated"
+directive @deprecated(
+    "The reason for the deprecation"
+    reason: String = "No longer supported"
+  ) on FIELD_DEFINITION | ENUM_VALUE
```
williamboman: Not quite sure why these started showing up in the SDL - https://github.com/graphql-java/graphql-java/pull/1723 maybe?

glasser:
See also #46 (these PRs should be looked at in parallel)

williamboman:
> See also #46 (these PRs should be looked at in parallel)

Thanks. Apart from bumping graphql-java this PR also introduces an (backwards-compatible) API for filtering directives in the SDL output: https://github.com/apollographql/federation-jvm/pull/41/commits/fed66babb7ce547c047ee1b9c36fd2ecd1ad38ed

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/FederationTest.java
```diff
@@ -23,29 +26,42 @@
 
 class FederationTest {
     private final String emptySDL = TestUtils.readResource("schemas/empty.graphql");
+    private final String emptySDLOutput = TestUtils.readResource("schemas/empty-output.graphql");
```
mcohen75: Refactoring this to a separate file is a nice change as this is more consistent with the rest of the tests here.

One thing to point out for other reviewers here is that the expected printed schema is different with this version because the include, skip and deprecated directives are now output.

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/FederationTest.java
```diff
@@ -23,29 +26,42 @@
...
     private final String interfacesSDL = TestUtils.readResource("schemas/interfaces.graphql");
     private final String isolatedSDL = TestUtils.readResource("schemas/isolated.graphql");
     private final String productSDL = TestUtils.readResource("schemas/product.graphql");
+    private final String productSDLOutput = TestUtils.readResource("schemas/product-output.graphql");
```
mcohen75: It wasn't clear to me at first why a separate schema is needed here. I realize now that it's because a few new directives started appearing and I suspect an error occurs when loading the schema with the directives expressed.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -45,6 +49,12 @@ public SchemaTransformer fetchEntities(DataFetcher entitiesDataFetcher) {
         return this;
     }
 
+    @NotNull
+    public SchemaTransformer directivePredicate(Predicate<GraphQLDirective> directivePredicate) {
```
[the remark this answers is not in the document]
williamboman: This change really deserves to be in a separate PR - but you can find more background [here](https://github.com/leangen/graphql-spqr/issues/290#issuecomment-562025005).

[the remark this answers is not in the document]
mcohen75: Yes I agree that it would be helpful for this to be in a separate PR because it's a separate bit of functionality.

If I'm following along correctly here it seems that the desire is to exclude the private directives because they cause the schema to fail to be transformed successfully. The failure is due to the fact that the directives are not included in the schema. Why is it desirable to use directives in a schema but not include the definition of the directives?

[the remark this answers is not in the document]
williamboman: > Why is it desirable to use directives in a schema but not include the definition of the directives?

I'm afraid I can't answer that question. I believe it's an implementation detail of graphql-spqr. Ping @kaqqao?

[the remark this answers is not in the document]
smyrick: This is actually part of the federation spec for the `sdl` field: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#query_service

We have to do the same logic here in graphql-kotlin-federation: https://github.com/ExpediaGroup/graphql-kotlin/blob/8ff7de80e8681b75091f244564a991aef7eea4fd/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt#L90-L97

I have created a PR in graphql-java to make this a little easier by adding a new option to the printer: https://github.com/graphql-java/graphql-java/pull/1756

kaqqao:
> Why is it desirable to use directives in a schema but not include the definition of the directives?

I don't fully understand how Apollo Federation works, so please excuse me if I say something incorrect. That said, I believe Apollo Federation operates on SDL and not introspection. The printed SDL might easily contain sensitive schema directives for e.g. access control that have no business ever being exposed. SPQR utilizes directives to store meta data about the fields and types. These are defined as schema directives that are never intended for the clients to be aware of (as opposed to client/query directives, that clients can provide as a part of the query and thus must be aware of). What would a client even do with the knowledge of such a directive (except perhaps abuse it)?

As of graphql-java v14, `SchemaPrinter` enables fine-grained filtering of directives that get printed in the SDL. I'd argue it is sensible to completely omit all schema directives (e.g. type or field level directives), as they're neither useful to the client (the client can't send them in a query or make use of them in any other way), nor are they normally accessible (they're never exposed via introspection). I am wondering why does Apollo even expect these to be printed. Additionally, if directives are used to store sensitive info like access control rules, exposing full SDL print outs becomes straight out dangerous practice.

graphql-java-support/src/test/resources/com/apollographql/federation/graphqljava/schemas/directives-output.graphql
sachindshinde: I believe it's from graphql-java/graphql-java#1609 , specifically the removal of
```
        // we don't print the standard directives that always ship with graphql-java
        List<String> standardDirectives = Arrays.asList(
                "skip",
                "include",
                "defer",
                "deprecated");
```

sachindshinde:
@kaqqao 

Sorry for the delays in resolving this issue. I talked to the team managing Apollo Gateway, and I have some answers here.

Regarding Apollo Federation, an important aspect to understand is that `federation-jvm` intentionally exposes "server-side" directives because a federated GraphQL server (a.k.a. an "implementing service") is intended to **not** be exposed to the client. From the [docs](https://www.apollographql.com/docs/apollo-server/federation/implementing-services/#securing-implementing-services):
>Because of the power and flexibility of Apollo Federation's `_entities` field, your implementing services should not be directly accessible by clients. Instead, only your gateway should have access to your implementing serivces. Clients then communicate with the gateway

Apollo Gateway, when composing the graphs of the implementing services, will effectively prevent server-side directives from being made visible to the client. More specifically, during the gateway's graph composition, the gateway will do the following:
- directive definitions that only contain type system locations will be completely dropped from the composed schema
- directive definitions that contain both type system locations and executable locations will still be in the composed schema, but will have their type system locations removed from the directive definition
- all directive usages on type system locations will be dropped, with the sole exception of the `@deprecated` directive

So while server-side directives are exposed via `federation-jvm`, they are not exposed to clients.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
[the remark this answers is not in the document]
sachindshinde: There's been a slight misunderstanding about Apollo Federation, there's more info at  https://github.com/apollographql/federation-jvm/pull/41#issuecomment-599715362

sachindshinde:
@williamboman 

Thanks for the PR! There's a few changes here that are desirable
- There doesn't appear to be an immediate need to expose `includeDirectives` functionality to users, due to the clarification in https://github.com/apollographql/federation-jvm/pull/41#issuecomment-599715362 .
- We shouldn't emit directive definitions for standard directives (`@deprecated`, `@include`, `@skip`) into the SDL, as Apollo Gateway graph composition will fail if they appear in SDL. Luckily, the `includeDirectives` option in graphql-java v14 does just this (it also doesn't filter out usages of `@deprecated`, which is critical).

I've went ahead and made a PR that does this at #51 .

elxnis:
`There doesn't appear to be an immediate need to expose includeDirectives functionality to users, due to the clarification in #41 (comment) .`
Hi, yes there is. Clarification in comment 41 holds only if you assume that server can't be accessed directly and only through gateway and that is not true for all usecases.
Furthermore there is issue about exposing directive definition and actual application.
When PR #51 is merged I can offer a small enhancement that gives more control to server owner what is exposed with federation-jvm extension, i.e. ability to define our own sdl data fetcher as it looks we are going in circles of how people what it to work and how apollo sees it being used.

smyrick:
This can be closed as the version was upgraded in https://github.com/apollographql/federation-jvm/pull/51

PR #43, opened 2020-01-26
Schemas' comparison should not take line separators into account

rkudryashov:


glasser:
Is this purely a test issue, or is there something else going on in the actual code to introduce odd line separators? eg, is this a symptom of checking out a git repo in some way that changes newlines in files in the repo?

rkudryashov:
@glasser yes, this is only a test issue. Tests use a method that returns different results depending on the OS used: https://github.com/apollographql/federation-jvm/blob/829aec8f864d4653d38fc20b68135081c85dab29/graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/TestUtils.java#L16
Explanation: http://www.cs.toronto.edu/~krueger/csc209h/tut/line-endings.html

glasser:
Would it make sense to change TestUtils.readResource to always use `\n` instead?

glasser:
(@pcarrier, do you remember why you made TestUtils.readResource convert line endings to the system's line ending?)

rkudryashov:
It would, I've changed implementation.

pcarrier:
Looks good to me. Not sure why I went with system separator.

glasser:
Can we just skip the split and rejoin then?

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/TestUtils.java
```diff
@@ -6,13 +6,13 @@
...
-    }
+
+    private static final String BASE_LINE_SEPARATOR = "\n";
 
     static String readResource(String name) {
         InputStream is = SchemaUtils.class.getResourceAsStream(name);
         assert is != null;
         BufferedReader reader = new BufferedReader(new InputStreamReader(is));
-        return reader.lines().collect(Collectors.joining(System.lineSeparator()));
+        return reader.lines().collect(Collectors.joining(BASE_LINE_SEPARATOR));
```
glasser: OK, so what this does is normalize any line endings on disk to unix newline. That seems reasonable.

PR #48, opened 2020-02-11
Cleanup

rkudryashov:
.idea folder is recreated for everyone automatically and shouldn't be under VCS.

After pull you may need to reimport project using IDEA, which will automatically create this folder.


rkudryashov:
Accidentally pushed another change with moving out String constants to a separate class, but maybe PR doesn't look huge so far.

rkudryashov:
We don't get benefits by holding static constants in different classes. Also, by convention they should be written in upper case.

smyrick:
This is an API change + a repository clean up. Would it be better to just make a small PR to add `.idea` to the gitignore then a separate, unrelated PR to add the `Constants` if needed

sachindshinde:
> .idea folder is recreated for everyone automatically and shouldn't be under VCS.
>
> After pull you may need to reimport project using IDEA, which will automatically create this folder.

The idea behind keeping the `.idea` folder in the repo is that we can use it to support certain style/formatting guidelines.

> We don't get benefits by holding static constants in different classes. Also, by convention they should be written in upper case.

I definitely agree with the uppercase convention, that should be followed here.

Regarding the benefits of holding static constants in different classes, the benefit is that we can keep constants related to a particular concept grouped together. If we placed all constants under one static class, we'd have to represent the hierarchy in the name (as you've currently done), which can get messy.

rkudryashov:
@smyrick 

Yeah, it would be better, but probably this PR will be closed.

rkudryashov:
@sachindshinde 

> The idea behind keeping the .idea folder in the repo is that we can use it to support certain style/formatting guidelines.

I understand this intention, but in my opinion, it would be better to use other means (such as `.editorconfig` file, common IDEA-settings file not under VCS, or other). Also, I have looked at some well-known Java repositories with much more contributors than at `federation-jvm`, for instance:

- https://github.com/graphql-java/graphql-java
- https://github.com/spring-projects/spring-boot
- https://github.com/junit-team/junit5
- https://github.com/apache/maven

and well, some of them (spring-boot in this case) do store `idea` folder. But it seems like this approach is used in minority of repositories.

On the static constants, I can notice that this is of course up to a library designer what approach to use. This PR was implied to be the first part of global refactoring; the following parts would allow getting rid of mapping Apollo entities to Java classes. I don't think `Constants` class can get messy due to the small number of constants (9 after applying all changes: https://github.com/rkudryashov/graphql-java-federation/blob/master/src/main/java/io/gqljf/federation/misc/Constants.java).

Feel free to close this PR and maybe use any part of it in the future PRs. Probably, I will continue to develop fork of this project: https://github.com/rkudryashov/graphql-java-federation (it already switched to GJ 14, contains just 9 classes, and has some important features such as the ability to disable subscriptions while working under Apollo Server due to this issue: https://github.com/apollographql/apollo-server/issues/3357).

PR #50, opened 2020-03-16
Update .idea directory and .gitignore

sachindshinde:
I'm guessing these are added from a newer IntelliJ


smyrick:
Isn't the better option to ignore the entire `.idea` folder and apply common styling with a `.editorconfig` file that supports multiple editors, not just IntelliJ?

https://editorconfig.org/

sachindshinde:
We ended up switching to Spotless code formatting and removing all `.idea` files from the repo in #124 , so this PR is no longer needed.

PR #51, opened 2020-03-16
graphql-java-support: Bump graphql-java to v14

sachindshinde:
This PR is an iteration of the changes suggested in #41 and #46 (many thanks to the authors and commenters).

Summary of changes:
- Bump `graphql-java` to V14 in `pom.xml`
- Change `GraphQLType::getName()` to `GraphQLNamedType::getName()`
  - Reason: `graphql-java` changed their type hierarchy
- Change `SchemaPrinter.Options.includeSchemaDefintion()` to `includeSchemaDefinition()`
  - Reason: `graphql-java` fixed a typo in the word "definition"
- Change `RuntimeWiring.getSchemaTransformers()` to `RuntimeWiring.getSchemaGeneratorPostProcessings()`
  - Reason: `graphql-java` changed the getter name
- Prevent standard directive definitions (`@deprecated`, `@include`, and `@skip`) from appearing in the SDL
  - Reason: `graphql-java` now includes these definitions in the schema printer by default (due to graphql-java/graphql-java#1609 ), which breaks gateway composition
- Update `FederationSdlPrinter` to use `graphql-java` v14's `SchemaPrinter`
  - Reason: As described in #52 , a bug in `graphql-java`'s `SchemaPrinter` forces us to maintain our own copy (`FederationSdlPrinter`), but with the bugs fixed. As described in #53 , we also use our own printer to avoid printing federation directive definitions and federation type definitions in SDL.

Note this is a backwards-incompatible change.

~~Also note that `FederationSdlPrinter.Options.includeDirectives()` will filter all directive definitions and usages indicated with the exception of directive usages of `@deprecated`, which is conveniently the behavior we want.~~ We've added directive definition filtering in #53 , so we don't need to rely on `FederationSdlPrinter.Options.includeDirectives()` to filter the standard directive definitions.

It'll be easier in this PR to go commit-by-commit. For `FederationSdlPrinter`, it'll help to compare `FederationSdlPrinter.java` in the class-copy-and-rename commit to the v14 `SchemaPrinter`, located at https://raw.githubusercontent.com/graphql-java/graphql-java/v14.0/src/main/java/graphql/schema/idl/SchemaPrinter.java . For the `FederationSdlPrinter` changes that fix the issues in #52 and #53 , it'll help to compare against the respective changes in those PRs.


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -142,8 +148,8 @@ private String sdl() {
         final SchemaPrinter.Options options = SchemaPrinter.Options.defaultOptions()
                 .includeScalarTypes(true)
                 .includeExtendedScalarTypes(true)
-                .includeSchemaDefintion(true)
-                .includeDirectives(true);
+                .includeSchemaDefinition(true)
+                .includeDirectives(directive -> !HIDDEN_DIRECTIVES.contains(directive.getName()));
```
mcohen75: This will exclude the `@deprecated` directive. This directive is an important part of GraphQL APIs as this is how deprecations are communicated to API consumers. At Indeed our strategy for evolving schema relies heavily on this directive. Are you certain that this breaks composition of schemas at the gateway? If so, maybe it's best to fix this in the gateway?

sachindshinde: This will only exclude the directive definitions of `@deprecated`, `@include`, and `@skip` from the SDL.

It will not remove usages of the `@deprecated` directive. This is visible in `SchemaPrinter`'s source code at [https://github.com/graphql-java/graphql-java/blob/v14.0/src/main/java/graphql/schema/idl/SchemaPrinter.java#L715](https://github.com/graphql-java/graphql-java/blob/v14.0/src/main/java/graphql/schema/idl/SchemaPrinter.java#L715) and [https://github.com/graphql-java/graphql-java/blob/v14.0/src/main/java/graphql/schema/idl/SchemaPrinter.java#L681](https://github.com/graphql-java/graphql-java/blob/v14.0/src/main/java/graphql/schema/idl/SchemaPrinter.java#L681) .

To be safe, I've updated `FederationTest` to include an `@deprecated` directive on a type so we can know if `SchemaPrinter` changes its behavior in the future.

Apollo Gateway will add back the directive definitions for `@deprecated`, `@include`, and `@skip` when it composes the graphs of the implementing services.

mcohen75: Ah yes that's correct, I was misled by the `includeDirectives` name.

mcohen75:
> Also note that SchemaPrinter.Options.includeDirectives() will filter all directive definitions and usages indicated with the exception of directive usages of @deprecated, which is conveniently the behavior we want.

Is it worth adding a test to ensure that the `@skip` and `@include` directives are filtered out? This would catch problems here in the future and would serve as convenient documentation when we all forget about this PR.

sachindshinde:
@mcohen75 
`@include` and `@skip` can only be used on executable locations (as per spec), and not type system locations. The only way we could test is by including executable definitions, but `SchemaParser.parse()` will naturally throw if it sees those.

graphql-java-support/src/test/java/com/apollographql/federation/graphqljava/SchemaUtils.java
```diff
@@ -3,20 +3,27 @@
...
 import java.util.Map;
+import java.util.Set;
 
 import static graphql.ExecutionInput.newExecutionInput;
 import static graphql.GraphQL.newGraphQL;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 final class SchemaUtils {
+    static final Set<String> standardDirectives =
```
zionts: minor: should we share this const val?

sachindshinde: We shouldn't, I'll make it `private`.

sachindshinde:
So after some testing in our own infra, I've opened up #53 to address some further issues, and rebased this branch on top of that.

The main consequences for this PR are that we can just use the directive definition filtering from #53 to hide standard directive definitions.

PR #58, opened 2020-04-07
Update valid locations for @extends

smyrick:
According to the spec `@extends` is valid for both OBJECT and INTERFACE: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#federation-schema-specification

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -11,6 +11,7 @@ import graphql.schema.GraphQLDirective;
 import graphql.schema.GraphQLNonNull;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


smyrick:
Note: Here are the change that were made when we migrate from our own implementation of federation to this library: https://github.com/ExpediaGroup/graphql-kotlin/pull/654

Ideally there would be no unit test changes

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -49,7 +50,7 @@
             .type(new GraphQLNonNull(_FieldSet.type))
             .build();
 
-    private static final GraphQLArgument fieldsArgument(String value) {
+    private static GraphQLArgument fieldsArgument(String value) {
```
smyrick, line 53: static methods don't need to be marked as final

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,10 @@
+<component name="InspectionProjectProfileManager">
```
smyrick, line 1: Ideally these files are just all deleted and ignored. But if you want me to manually exclude these changes please let me know

smyrick:
@sachindshinde All requested changes have been merged in. I think the updates to `encodings.xml` are valid as IntelliJ added those for me.

Issue #71, opened 2020-06-17
Quotes within docstrings result in unparsable schema from _Service.sdl


fotoetienne:
Example:

```graphql
type Query {
    """Field docstring contains a "quoted" field."""
    fooField: String
}
```

When queried from sdl becomes

```graphql
schema {
  query: Query
}

type Query {
 "Field docstring contains a "quoted" field."
  fooField: String
}
```

The description string should be surrounded in triple quotes as in the original schema.

fotoetienne:
Looks like the problem is [here](https://github.com/apollographql/federation-jvm/blob/master/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java#L936). For single-line descriptions, we automatically use single-quotes. We should force triple-quotes if there are any characters that can't be contained in a graphql [StringValue](https://spec.graphql.org/draft/#sec-String-Value) (`"`, `\`, or `LineTerminator`)

sachindshinde:
Thanks for reporting this bug! We've merged a fix in, which has been released in v0.4.2.

PR #72, opened 2020-06-19
Render descriptions as BlockStrings if necessary

fotoetienne:
Currently, quotes within decription strings result in unparseable schema from _Service.sdl

Example:

```graphql
type Query {
    """Field docstring contains a "quoted" field."""
    fooField: String
}
```

When queried from sdl becomes

```graphql
schema {
  query: Query
}

type Query {
 "Field docstring contains a "quoted" field."
  fooField: String
}
```

In this case, the description string should be surrounded in triple quotes as in the original schema.
We should force triple-quotes if there are any characters that can't be contained in a graphql [StringValue](https://spec.graphql.org/draft/#sec-String-Value) (`"`, `\`, or `LineTerminator`)

See also https://github.com/apollographql/federation-jvm/issues/71

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
@@ -932,14 +932,16 @@ public class FederationSdlPrinter {
             List<String> lines = Arrays.asList(descriptionText.split("\n"));
             if (options.isDescriptionsAsHashComments()) {
                 printMultiLineHashDescription(out, prefix, lines);
+            } else if (needsBlockString(descriptionText)) {
+                printMultiLineDescription(out, prefix, lines);
             } else {
-                if (lines.size() > 1) {
-                    printMultiLineDescription(out, prefix, lines);
-                } else {
-                    printSingleLineDescription(out, prefix, lines.get(0));
-                }
+                printSingleLineDescription(out, prefix, descriptionText);
             }
         }
+    }
+
+    private boolean needsBlockString(String description) {
+        return description.contains("\"") || description.contains("\n") | description.contains("\r");
     }
 
     private void printMultiLineHashDescription(PrintWriter out, String prefix, List<String> lines) {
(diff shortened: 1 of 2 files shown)


sachindshinde:
Thanks for the PR! `FederationSdlPrinter` is actually a copy of the `SchemaPrinter` class from `graphql-java`, with some tweaks to allow us to print federation SDL. It looks like this bug is from `graphql-java` that was fixed more generally in graphql-java/graphql-java#1832 , which was released in `graphql-java` v15.

We'll be upgrading `federation-jvm` to support `graphql-java` v15 shortly, but in the meanwhile we can backport the fix from graphql-java/graphql-java#1832 to our current `FederationSdlPrinter` for users on v14. I'll file a PR for backport shortly.

sachindshinde:
I've filed PR #74 for the backport, and added a test which encompasses your use case.

graphql-java-support/src/test/resources/com/apollographql/federation/graphqljava/schemas/product.graphql
```diff
@@ -8,7 +8,11 @@ type Money {
 }
 
 type Product @key(fields : "upc") {
+  """
+  The product "name".
+  """
```
mcohen75, line 13: Would it make sense to include test cases for all characters that require block strings here?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
```diff
@@ -932,16 +932,18 @@ private void printComments(PrintWriter out, Object graphQLType, String prefix) {
...
-                } else {
-                    printSingleLineDescription(out, prefix, lines.get(0));
-                }
+                printSingleLineDescription(out, prefix, descriptionText);
             }
         }
     }
 
+    private boolean needsBlockString(String description) {
+        return description.contains("\"") || description.contains("\n") | description.contains("\r");
```
mcohen75, line 944: According to the spec a backslash is also a character that can be used in Block Strings without escaping. At Indeed we use line feed and quote frequently, so this definitely hits the main ones we care about.

mcohen75: See http://spec.graphql.org/draft/#sec-String-Value.Block-Strings

PR #67, opened 2020-06-03
Upgrade to graphql-java 15

smyrick:
See the release notes: https://github.com/graphql-java/graphql-java/releases/tag/v15.0

* Includes new directive in the spec `@specifiedBy`
* Removed extended scalars from the schema
* Included bug fix https://github.com/graphql-java/graphql-java/pull/1798

PR #52, opened 2020-03-18
graphql-java-support: Fix grammar violation in graphql-java's SchemaPrinter

sachindshinde:
`graphql-java`'s `SchemaPrinter` outputs grammar-violating SDL, as described in graphql-java/graphql-java#1823 . This has affected users of `federation-jvm`, as shown in #21 and #23 .

It's been fixed upstream by graphql-java/graphql-java#1798 , but this hasn't been released yet. This PR introduces a hotfix until `graphql-java` releases their fix (and backports it to `graphql-java` v13).

The gist is we copy `SchemaPrinter` from `graphql-java` v13 as `FederationSdlPrinter`, and backport the fixes from graphql-java/graphql-java#1798 . I've added a basic test to `FederationTest` to check behavior for the relevant types in the backfix (objects, interfaces, enums, and input objects).

For reviewing, it'll help to diff `FederationSdlPrinter` against the v13 `SchemaPrinter`, located at https://raw.githubusercontent.com/graphql-java/graphql-java/v13.0/src/main/java/graphql/schema/idl/SchemaPrinter.java , and compare the diff against the changes files at graphql-java/graphql-java#1798 .

PR #73, opened 2020-06-27
graphql-java-support: Bump graphql-java to v15

sachindshinde:
This PR is an iteration of the changes suggested in #67 (many thanks to the authors and commenters).

Summary of Changes:
- Bump `graphql-java` to v15 in the `pom.xml`
- Remove `enforceSchemaDirectives(true)` from `Federation.java`
  - Reason: This option was removed in graphql-java/graphql-java#1773 to facilitate spec compliance, and now is always `true` (thankfully we've already done the legwork to not have to worry about this)
- Remove `includeExtendedScalarTypes(true)` from `SchemaTransformer.java`
  - Reason: This option was removed in graphql-java/graphql-java#1780 to facilitate spec compliance, and now is always `true` (what were previously known as "extended scalars" in `graphql-java` are now just regular custom scalars)
- Add a new standard GraphQL directive `@specifiedBy` to standard directive lists
  - Reason: This directive was added to `graphql-java` in graphql-java/graphql-java#1772 with the spec addition still pending at graphql/graphql-spec#635
  - Note that it's unknown how gateway currently responds to the presence of this directive definition in the federation SDL, but it's assumed that it will eventually behave the same as it does with other standard directive definitions if it doesn't already do so. (This may introduce an issue in the interim that a user wanting to use `@specifiedBy` with federation will not be able to do so, since federation SDL won't expose the directive definition and the directive definitions the gateway introduces may not have it.)
- Update `FederationSdlPrinter` to use `graphql-java` v15's `SchemaPrinter`
  - Note that the bugfix that we backported in #52 is no longer needed with graphql-java v15.

Note that this is a backwards incompatible change due to backwards incompatible changes in `graphql-java` v15's `SchemaPrinter`.

It'll be easier in this PR to go commit-by-commit. For `FederationSdlPrinter`, it'll help to compare `FederationSdlPrinter.java` in the class-copy-and-rename commit to the v15 `SchemaPrinter`, located at https://raw.githubusercontent.com/graphql-java/graphql-java/v15.0/src/main/java/graphql/schema/idl/SchemaPrinter.java . For the `FederationSdlPrinter` changes that fix the issue in #53 , it'll help to compare against the respective changes in that PR.


sachindshinde:
Note that we should test this in our own Apollo infra before releasing

.idea/misc.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="JavaScriptSettings">
```
pcarrier, line 4: ?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
```diff
@@ -462,17 +468,36 @@ private void printFieldDefinitions(PrintWriter out, Comparator<? super GraphQLSc
...
             if (shouldPrintAsAst(type.getDefinition())) {
                 printAsAst(out, type.getDefinition(), type.getExtensionDefinitions());
             } else {
                 printComments(out, type, "");
-                out.format("interface %s%s", type.getName(), directivesString(GraphQLInterfaceType.class, type.getDirectives()));
+                if (type.getInterfaces().isEmpty()) {
+                    out.format("interface %s%s", type.getName(), directivesString(GraphQLInterfaceType.class, type.getDirectives()));
+                } else {
+
+                    GraphqlTypeComparatorEnvironment environment = GraphqlTypeComparatorEnvironment.newEnvironment()
```
pcarrier, line 479: Looks static

.idea/misc.xml
sachindshinde: My IntelliJ's been adding that in for a while, no clue why it keeps doing that (maybe it's because of plugins?).

PR #92, opened 2021-02-11
fix key directive should be repeatable

tinnou:
## Description

When using multiple @key directives on a type, an [AssertException](https://github.com/graphql-java/graphql-java/blob/b01df7449bebef5bd645f0255a94d8eaacfa0a5c/src/main/java/graphql/DirectivesUtil.java#L146) (this was a graphql-java 16 change) is triggered in the SchemaTransformer since the key directive is defined to be non repeatable. 

## Fix & Testing

The fix consists of switching the key directive to be repeatable as well as using `GraphQLDirectiveContainer.getDirectives()` instead of `GraphQLDirectiveContainer.getDirective()` (singular) reserved for non repeatable directives only.

I added an additional @key in one of the schemas in the tests to reproduce the issue.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -61,7 +61,7 @@ public final class FederationDirectives {
             .type(new NonNullType(new TypeName(_FieldSet.typeName)))
             .build();
 
-    /* directive @key(fields: _FieldSet!) on OBJECT | INTERFACE */
+    /* directive @key(fields: _FieldSet!) repeatable on OBJECT | INTERFACE */
 
     public static final String keyName = "key";
 
@@ -75,7 +75,6 @@ public final class FederationDirectives {
     public static GraphQLDirective key(String fields) {
         return newDirective(key)
                 .argument(fieldsArgument(fields))
-                .repeatable(true)
                 .build();
     }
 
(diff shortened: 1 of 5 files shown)


sachindshinde:
Thanks for noticing the issue and submitting a fix! I'll review this sometime early this week so we can ship the fix soon 🚀

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -69,18 +69,21 @@ private static GraphQLArgument fieldsArgument(String value) {
             .name(keyName)
             .validLocations(OBJECT, INTERFACE)
             .argument(fieldsArgument)
+            .repeatable(true)
```
sachindshinde, line 72: In the comment of
```java
/* directive @key(fields: _FieldSet!) on OBJECT | INTERFACE */
```
above, can you add `repeatable` there?

sachindshinde (changes requested):
PR's looking good, just some minor changes.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
tinnou: ok will do!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -69,18 +69,21 @@ private static GraphQLArgument fieldsArgument(String value) {
...
             .build();
 
     public static GraphQLDirective key(String fields) {
         return newDirective(key)
                 .argument(fieldsArgument(fields))
+                .repeatable(true)
```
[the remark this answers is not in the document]
tinnou: ah, good catch.

sachindshinde (approved):
LGTM

PR #112, opened 2021-06-22
Update usage of AstPrinter to new graphql-java signature.

paulbakker:
There are some changes in graphql-java that affect the `FederationSdlPrinter`: https://github.com/graphql-java/graphql-java/commit/3139ccbb8a2313704518f39150223bda550d0500

This PR updates to the new method signatures.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
@@ -2,8 +2,8 @@ package com.apollographql.federation.graphqljava;
 
 import graphql.Assert;
 import graphql.PublicApi;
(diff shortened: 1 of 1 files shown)


paulbakker:
Build fails because it tries building against the released version of graphql-java. Not sure what the workflow is to build against the latest source of graphql-java for this project. 

Locally I built against a snapshot, but that's probably not what we want here.

sachindshinde:
@paulbakker 
Thanks for filing the PR! It's good to know about the upcoming breaking changes in `graphql-java` v17 and how we'd need to update our printer fork.

More generally we don't update the repo to build against the latest source of dependencies like `graphql-java`, but this PR will be useful once v17 is released.

paulbakker:
Closing this PR because another (newer) PR is available for this issue: https://github.com/apollographql/federation-jvm/pull/122

PR #115, opened 2021-07-03
Extract interfaces meant for integrations to api artifact

### LINKED ISSUE #114: Extract interfaces meant for integrations to api artifact
Extract HTTPRequestHeaders interface (only API integration interface right now) to its own artifact:

Benefits:

- Any framework that chooses to implement HTTPRequestHeaders shouldn't be required to pull in the remainder of all the transitive libraries this project depends upon.

- My company has protobuf as a banned library, and so we are required to shade protobuf and federation-jvm libraries in order to get federated tracing functioning.  By separating interfaces to a separate api artifact, we can shade federation-jvm, but leave the HTTPRequestHeaders interface in its original location.  This allows other 3rd party libraries we use to continue to implement HTTPRequestHeaders, and allows the optimization logic in inspecting the headers to still function while allowing those 3rd party libraries remain decoupled/insulated from the shading concern.

theJC:
Resolves #114 

This extracts any interfaces (only one right now) to a separate maven submodule so that projects and libraries that leverage HTTPRequestHeaders aren't required to pull in all of the transitive libraries that federation-jvm does.

Additionally as well, support the scenario when federation-jvm needs to be shaded, this interface needs to remain in place to allow integrations that implement this interface to still properly be honored to short circuit tracing when the header is not present.


sachindshinde:
Thanks for the contribution!

Agreed that integrations in third-party libraries shouldn't have to pull in all transitive dependencies when they're implementing APIs, so this package split makes sense to me.

There's some slight tweaks I'd like to make to the description and plugins in the `pom.xml`, but I can make those in a follow-up PR.

theJC:
FYI this was my first pass for determining how I was going to attempt to integrate federated instrumentation in some of our projects that leverage DGS.  

However https://github.com/apollographql/federation-jvm/pull/116 is a liberating change in that removes the onus from a the library implementing the graphql context from having to implementing the HTTPRequestHeaders interface and allows other strategies, allowing consuming codebases to control the responsibility of that integration, ie: of pulling in federation-jvm and DGS and being able to do their own necessary wiring.

PR #116, opened 2021-07-06
Provide alternative integration interface for FederatedTracingInstrum…

theJC:
…entation to access incoming http headers

Rather than requiring consumers to implement the HTTPRequestHeaders, this allows a consumer to construct FederatedTracingInstrumentation passing in via the options its own implementation on interrogating the context on the execution input (think Netflix DGS's or some other framework having its own standard context shape) to determine if it should trace or not (and leverage its own context implementation to interrogate the header).

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
@@ -37,6 +37,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import static graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted;
 import static graphql.schema.GraphQLTypeUtil.simplePrint;
@@ -60,24 +61,16 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
 
     @Override
     public InstrumentationState createState(InstrumentationCreateStateParameters parameters) {
-        // If we've been configured with a way of reading HTTP headers, we should only be active
-        // if the special HTTP header has the special value. If the header isn't provided or has
-        // a different value, return null for our state, which we'll interpret in the rest of this
-        // file as meaning "don't instrument".  (If we haven't been given access to HTTP headers,
-        // always instrument.)
         Object context = parameters.getExecutionInput().getContext();
-        if (context instanceof HTTPRequestHeaders) {
-            @Nullable String headerValue = ((HTTPRequestHeaders) context).getHTTPRequestHeader(HEADER_NAME);
-            if (headerValue == null || !headerValue.equals(HEADER_VALUE)) {
-                return null;
-            }
+        if (options.contextIndicatesWeShouldTrace(context)) {
+            return new FederatedTracingState();
         }
-        return new FederatedTracingState();
+        return null;
     }
 
     @Override
     public CompletableFuture<ExecutionResult> instrumentExecutionResult(ExecutionResult executionResult, InstrumentationExecutionParameters parameters) {
-        final @Nullable FederatedTracingState state = parameters.<FederatedTracingState>getInstrumentationState();
+        final @Nullable FederatedTracingState state = parameters.getInstrumentationState();
         if (state == null) {
             return super.instrumentExecutionResult(executionResult, parameters);
         }
@@ -178,7 +171,7 @@ public class FederatedTracingInstrumentation extends SimpleInstrumentation {
         }
 
         if (result instanceof DataFetcherResult<?>) {
-            DataFetcherResult<?> theResult = (DataFetcherResult) result;
+            DataFetcherResult<?> theResult = (DataFetcherResult<?>) result;
             if (theResult.hasErrors()) {
                 graphQLErrors.addAll(theResult.getErrors());
             }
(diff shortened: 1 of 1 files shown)


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -37,14 +37,16 @@
...
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import static graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted;
 import static graphql.schema.GraphQLTypeUtil.simplePrint;
 
 public class FederatedTracingInstrumentation extends SimpleInstrumentation {
+    public static final String HEADER_NAME = "apollo-federation-include-trace";
+    public static final String HEADER_VALUE = "ftv1";
```
theJC, line 47: Made public so integrators who provide their own evaluateContextFunction can leverage the same header name and value in their implementation.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -433,17 +429,39 @@ public Node(Reports.Trace.Node.Builder builder) {
...
+                // If context implements our HTTPRequestHeaders interface, we should only be active
+                // if the special HTTP header has the special value. If the header isn't provided or has
+                // a different value, return false - which is interpreted as meaning "don't instrument".
+                // (If context doesn't implement HTTPRequestHeaders, always instrument.)
+                 (context) -> {
+                     if (context instanceof HTTPRequestHeaders) {
+                        String header = ((HTTPRequestHeaders) context).getHTTPRequestHeader(HEADER_NAME);
+                        return HEADER_VALUE.equals(header);
+                    }
+                    return true;
```
theJC, line 451: Original implementation intact as the default, just slightly altered due to new way the question is being asked.

README.md
```diff
@@ -84,3 +84,25 @@ HTTPRequestHeaders context = new HTTPRequestHeaders() {
...
+takes the `context` object and returns a boolean indicating whether a trace should be created for
+the request being processed.  For example:
+
+```java
+Options options = new Options(false, (Object context) -> {
+    if (context instanceof MySpecialExecutionInputContext) {
+        // interrogate your context object... test for the existance of the headers and its value
+        // if header apollo-federation-include-trace exists and is set to ftv1
+        return true;
+        }
```
bono007, line 100: **[TinyNit]** Indentation

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
bono007: May want to consider naming this a bit more descriptive if opening up for public use. An alternative would be to provide static accessor methods for the header name and value so that consumers lock onto that rather than a const.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -433,9 +429,27 @@ public Node(Reports.Trace.Node.Builder builder) {
 
     public static class Options {
         private final boolean debuggingEnabled;
+        private final Predicate<Object> contextSignalsToTrace;
```
bono007, line 432: Wdyt about passing in the `ExecutionInput` rather than the `context`? Would implementors ever want to consult something besides the context - possibly the `GraphQLContext`? 

It would also have the benefit of the input being typed rather than `Object`.

Wdyt about naming it something more generic as well - like `shouldTrace`?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -433,9 +429,27 @@ public Node(Reports.Trace.Node.Builder builder) {
...
 
-        public Options(boolean debuggingEnabled) {
+        public Options(boolean debuggingEnabled, Predicate<Object> contextSignalsToTrace) {
```
bono007, line 434: Oh no - 2 parameters. The 3rd would definitely call for a builder. :)

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
theJC: Awesome suggestions!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java
```diff
@@ -433,9 +429,30 @@ public Node(Reports.Trace.Node.Builder builder) {
...
+                    (executionInput) -> {
+                        if (executionInput != null && executionInput.getContext() != null) {
+                            Object context = executionInput.getContext();
+                            if (context instanceof HTTPRequestHeaders) {
+                                String header = ((HTTPRequestHeaders) context).getHTTPRequestHeader(FEDERATED_TRACING_HEADER_NAME);
+                                return FEDERATED_TRACING_HEADER_VALUE.equals(header);
+                            }
+                        }
+                        return true;
+                    });
```
theJC, line 455: NOTE: Original implementation as the previous version, only slightly altered due to new way the question is being asked.

sachindshinde (approved):
LGTM! There's some slight changes I'd like to make to `Options` for nullability, but I can file those in a follow-up PR.

Issue #131, opened 2021-09-24
Move build system to Gradle?


martinbonnin:
Any thoughts about moving to Gradle for the build system? One immediate reason for doing so would be to use [in-memory GPG keys](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys) in CI and automating the release process. It might be doable with Maven too but it looks more complicated (need to store a cryped keystore in the repo or restore it from env variables which both do not sound really fun things to do).

Overall, Gradle looks more flexible and more in-line with tools we use for other projects like [apollo-android](https://github.com/apollographql/apollo-android) but maybe there are reasons to stick with Maven?

sachindshinde:
>Any thoughts about moving to Gradle for the build system?

I'm not strongly against Gradle, although I would note we have a lot of plugins and checks in our [Parent POM](https://github.com/apollographql/federation-jvm/blob/master/pom.xml), e.g. Spotless and several Maven Enforcer rules.

If we move, we should make sure we have equivalent checks and functionality in Gradle. It looks like there's a[ port of Maven Enforcer to Gradle](https://kordamp.org/enforcer-gradle-plugin/), but looking at the [feature-parity chart](https://kordamp.org/enforcer-gradle-plugin/#_comparison_to_maven), it's missing some rather important rules.

One of the important ones in our case is [RequireUpperBoundDeps](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html). This is a rule that checks that the dependency version resolved by Maven is the same as the max version among all transitive dependency declarations. This is useful for ensuring that Maven and Gradle have the same behavior during dependency version mediation/resolution (and also for ensuring that dependency exclusions don't accidentally exclude a higher version).

>One immediate reason for doing so would be to use in-memory GPG keys in CI and automating the release process. It might be doable with Maven too but it looks more complicated (need to store a cryped keystore in the repo or restore it from env variables which both do not sound really fun things to do).

From https://www.cloudbees.com/blog/continuous-deployment-maven-central-apache-maven it looks we can use the `pgp-maven-plugin` to pass them in through environment variables.

>Overall, Gradle looks more flexible and more in-line with tools we use for other projects like apollo-android but maybe there are reasons to stick with Maven?

[...]

PR #132, opened 2021-09-24
Add a helper script for releases

martinbonnin:
Add a `tag.main.kts $version` script that takes care of:

- setting versions in all pom.xml to `"$version"`
- setting the version in README.md
- creating a tag
- setting versions in all pom.xml to `"${version + 1}-SNAPSHOT"`

A next step would be to automate the release process completely by publishing releases and snapshots from CI (see also https://github.com/apollographql/federation-jvm/issues/131)

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -15,15 +15,13 @@
 - Delete the `private.key` file.
 
 ## For each release
-- Start a branch `release-VERSION`.
-- Update RELEASE_NOTES.md.
-- Edit all instances of (next version)-SNAPSHOT in all pom.xml files to be the desired version.
-- Edit version in Gradle section of README.md.
-- Push branch, open PR, wait for CI to pass.
-- Ensure you are using the latest Zulu build of OpenJDK 8.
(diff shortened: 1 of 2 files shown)


sachindshinde:
Sorry, I should have clarified here during our pairing sessions that there was a maven plugin we could do to do this, but we just hadn't looked into how to use it :bow: .

The plugin here is the [Maven Release Plugin](https://maven.apache.org/maven-release/maven-release-plugin/). It looks like [the steps](https://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html) in `mvn release:prepare` does everything besides update the version in the `README.md`. And from [this SO post](https://stackoverflow.com/questions/34629326/update-versions-contained-in-readme-on-maven-release), it looks like there's some approaches for auto-updating the `README.md` with project version. (Although if those don't work, I'd be fine with scripting that part.)

martinbonnin:
No problem at all! This was made with Gradle in mind (see https://github.com/apollographql/federation-jvm/issues/131) and mostly copypasta'd from https://github.com/apollographql/apollo-android/blob/dev-3.x/scripts/tag.main.kts.

Closing this for now. We can revisit when we settle on #131

PR #140, opened 2022-01-29
update readme with tracing using GraphQLContext map

dariuszkuc:
Updating README with information on how to use `FederatedTracingInstrumentation` together with `GraphQLContext` map.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,10 +72,13 @@ GraphQL graphql = GraphQL.newGraphQL(graphQLSchema)
 ```
 
 It is generally desired to only create traces for requests that actually come from Apollo Gateway, as they aren't
-helpful if you're connecting directly to your backend service for testing. In order
-for `FederatedTracingInstrumentation` to know if the request is coming from Gateway, you need to give it access to the
-HTTP request's headers, by making the `context` part of your `ExecutionInput` implement the `HTTPRequestHeaders`
-interface. For example:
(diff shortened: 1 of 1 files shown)


martinbonnin:
Thanks for sending this! Looks good to me. More general question/observation: are there use cases that require "Custom GraphQL context" ? I was under the impression that everything was possible using `GraphQLContext` and it's a more future proof way to do it ?

dariuszkuc:
@martinbonnin besides support for "legacy" way as some apps may not be able to quickly update, I don't think there are any reasons to continue using custom context. I am planning to drop support for legacy deprecated context in the next major version of `graphql-kotlin`.

martinbonnin:
In that case, I'd happily remove "Custom GraphQL context" from the docs? New users can use `GraphQLContext` and legacy users can keep using "Custom GraphQL context" but no need to advertise it much? @sachindshinde any thoughts?

sachindshinde:
@martinbonnin That sounds fine to me (especially since there's no reasons for new users to use the deprecated custom GraphQL context).

martinbonnin:
@dariuszkuc is that ok with you? If yes, do you mind removing the section alltogether?

dariuszkuc:
Dropped deprecated context usage from readme.

martinbonnin (approved):

Issue #143, opened 2022-03-10
Federation 2 support


martinbonnin:
After discussion with @BoD and @pcmanus, the initial step is to add [definitions for `@link` and `@shareable` ](https://gist.github.com/pcmanus/46a9f81b9edb4768e6f322b3242dc9b1) in the returned schema. 

Support for directives rename is out of scope at the moment

hwillson:
This was completed in https://github.com/apollographql/federation-jvm/pull/144 🎉

PR #144, opened 2022-03-11
Add the Fed2 directives

martinbonnin:
See https://github.com/apollographql/federation-jvm/issues/143
See also https://github.com/apollographql/supergraph-demo-fed2/pull/61 for usage in a real life environment

### DIFF
--- a/graphql-java-support/pom.xml
+++ b/graphql-java-support/pom.xml
@@ -70,7 +70,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <protocArtifact>com.google.protobuf:protoc:3.9.0:exe:${os.detected.classifier}</protocArtifact>
+                    <protocArtifact>com.google.protobuf:protoc:3.19.4:exe:${os.detected.classifier}</protocArtifact>
                 </configuration>
             </plugin>
             <plugin>
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
@@ -136,12 +136,21 @@ public final class Federation {
       typeRegistry.add(_FieldSet.definition);
     }
 
-    // Also add the implementation for _FieldSet.
-    if (!runtimeWiring.getScalars().containsKey(_FieldSet.typeName)) {
-      return copyRuntimeWiring(runtimeWiring).scalar(_FieldSet.type).build();
-    } else {
-      return runtimeWiring;
+    // Add scalar type for link__Import, since the directives depend on it.
+    if (!typeRegistry.getType(link__Import.typeName).isPresent()) {
+      typeRegistry.add(link__Import.definition);
     }
+
+    // Also add the implementation for _FieldSet.
+    RuntimeWiring newRuntimeWiring = runtimeWiring;
+    if (!runtimeWiring.getScalars().containsKey(_FieldSet.typeName)) {
+      newRuntimeWiring = copyRuntimeWiring(newRuntimeWiring).scalar(_FieldSet.type).build();
+    }
+    if (!runtimeWiring.getScalars().containsKey(link__Import.typeName)) {
+      newRuntimeWiring = copyRuntimeWiring(newRuntimeWiring).scalar(link__Import.type).build();
+    }
+
+    return newRuntimeWiring;
   }
 
   private static RuntimeWiring.Builder copyRuntimeWiring(RuntimeWiring runtimeWiring) {
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
@@ -12,7 +12,6 @@ import java.io.Reader;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
-
 import org.jetbrains.annotations.NotNull;
 
 public final class Federation {
@@ -130,10 +129,12 @@ public final class Federation {
     // Add Federation directives if they don't exist.
     Stream<DirectiveDefinition> directivesToAdd = FederationDirectives.allDefinitions.stream();
     if (isFederation2) {
-      directivesToAdd = Stream.concat(directivesToAdd, FederationDirectives.federation2Definitions.stream());
+      directivesToAdd =
+          Stream.concat(directivesToAdd, FederationDirectives.federation2Definitions.stream());
     }
 
-    directivesToAdd.filter(def -> !typeRegistry.getDirectiveDefinition(def.getName()).isPresent())
+    directivesToAdd
+        .filter(def -> !typeRegistry.getDirectiveDefinition(def.getName()).isPresent())
         .forEachOrdered(typeRegistry::add);
 
     // Add scalar type for _FieldSet, since the directives depend on it.
@@ -202,17 +203,23 @@ public final class Federation {
   }
 
   public static boolean isFederation2(TypeDefinitionRegistry typeDefinitionRegistry) {
-    return typeDefinitionRegistry.getSchemaExtensionDefinitions().stream().anyMatch(
-        schemaExtensionDefinition -> schemaExtensionDefinition.getDirectives().stream().anyMatch(
-            directive -> directive.getName().equals("link") && directive.getArguments().stream().anyMatch(
-                argument -> {
-                  Value value = argument.getValue();
-                  return  argument.getName().equals("url")
-                      && value instanceof StringValue
-                      && ((StringValue)value).getValue().equals("https://specs.apollo.dev/federation/v2.0");
-                }
-            )
-        )
-    );
+    return typeDefinitionRegistry.getSchemaExtensionDefinitions().stream()
+        .anyMatch(
+            schemaExtensionDefinition ->
+                schemaExtensionDefinition.getDirectives().stream()
+                    .anyMatch(
+                        directive ->
+                            directive.getName().equals("link")
+                                && directive.getArguments().stream()
+                                    .anyMatch(
+                                        argument -> {
+                                          Value value = argument.getValue();
+                                          return argument.getName().equals("url")
+                                              && value instanceof StringValue
+                                              && ((StringValue) value)
+                                                  .getValue()
+                                                  .equals(
+                                                      "https://specs.apollo.dev/federation/v2.0");
+                                        })));
   }
 }
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
@@ -1,9 +1,6 @@
 package com.apollographql.federation.graphqljava;
 
-import graphql.language.FieldDefinition;
-import graphql.language.ObjectTypeDefinition;
-import graphql.language.TypeDefinition;
-import graphql.language.TypeName;
+import graphql.language.*;
 import graphql.schema.GraphQLSchema;
 import graphql.schema.idl.RuntimeWiring;
 import graphql.schema.idl.SchemaGenerator;
@@ -14,6 +11,8 @@ import java.io.File;
 import java.io.Reader;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
+
 import org.jetbrains.annotations.NotNull;
 
 public final class Federation {
@@ -43,12 +42,14 @@ public final class Federation {
   public static SchemaTransformer transform(
       final TypeDefinitionRegistry typeRegistry, final RuntimeWiring runtimeWiring) {
     final boolean queryTypeShouldBeEmpty = ensureQueryTypeExists(typeRegistry);
+
+    boolean isFederation2 = isFederation2(typeRegistry);
     RuntimeWiring newRuntimeWiring =
-        ensureFederationDirectiveDefinitionsExist(typeRegistry, runtimeWiring);
+        ensureFederationDirectiveDefinitionsExist(typeRegistry, runtimeWiring, isFederation2);
     final GraphQLSchema original =
         new SchemaGenerator()
             .makeExecutableSchema(generatorOptions, typeRegistry, newRuntimeWiring);
-    return transform(original, queryTypeShouldBeEmpty);
+    return transform(original, queryTypeShouldBeEmpty).setFederation2(isFederation2);
   }
 
   public static SchemaTransformer transform(final TypeDefinitionRegistry typeRegistry) {
@@ -125,10 +126,14 @@ public final class Federation {
   }
 
   private static RuntimeWiring ensureFederationDirectiveDefinitionsExist(
-      TypeDefinitionRegistry typeRegistry, RuntimeWiring runtimeWiring) {
+      TypeDefinitionRegistry typeRegistry, RuntimeWiring runtimeWiring, boolean isFederation2) {
     // Add Federation directives if they don't exist.
-    FederationDirectives.allDefinitions.stream()
-        .filter(def -> !typeRegistry.getDirectiveDefinition(def.getName()).isPresent())
+    Stream<DirectiveDefinition> directivesToAdd = FederationDirectives.allDefinitions.stream();
+    if (isFederation2) {
+      directivesToAdd = Stream.concat(directivesToAdd, FederationDirectives.federation2Definitions.stream());
+    }
+
+    directivesToAdd.filter(def -> !typeRegistry.getDirectiveDefinition(def.getName()).isPresent())
         .forEachOrdered(typeRegistry::add);
 
     // Add scalar type for _FieldSet, since the directives depend on it.
@@ -136,18 +141,20 @@ public final class Federation {
       typeRegistry.add(_FieldSet.definition);
     }
 
-    // Add scalar type for link__Import, since the directives depend on it.
-    if (!typeRegistry.getType(link__Import.typeName).isPresent()) {
-      typeRegistry.add(link__Import.definition);
-    }
-
     // Also add the implementation for _FieldSet.
     RuntimeWiring newRuntimeWiring = runtimeWiring;
     if (!runtimeWiring.getScalars().containsKey(_FieldSet.typeName)) {
       newRuntimeWiring = copyRuntimeWiring(newRuntimeWiring).scalar(_FieldSet.type).build();
     }
-    if (!runtimeWiring.getScalars().containsKey(link__Import.typeName)) {
-      newRuntimeWiring = copyRuntimeWiring(newRuntimeWiring).scalar(link__Import.type).build();
+
+    // Add scalar type for link__Import, since the directives depend on it.
+    if (isFederation2) {
+      if (!typeRegistry.getType(link__Import.typeName).isPresent()) {
+        typeRegistry.add(link__Import.definition);
+      }
+      if (!runtimeWiring.getScalars().containsKey(link__Import.typeName)) {
+        newRuntimeWiring = copyRuntimeWiring(newRuntimeWiring).scalar(link__Import.type).build();
+      }
     }
 
     return newRuntimeWiring;
@@ -192,5 +199,20 @@ public final class Federation {
     runtimeWiring.getSchemaGeneratorPostProcessings().forEach(builder::transformer);
 
     return builder;
+  }
+
+  public static boolean isFederation2(TypeDefinitionRegistry typeDefinitionRegistry) {
+    return typeDefinitionRegistry.getSchemaExtensionDefinitions().stream().anyMatch(
+        schemaExtensionDefinition -> schemaExtensionDefinition.getDirectives().stream().anyMatch(
+            directive -> directive.getName().equals("link") && directive.getArguments().stream().anyMatch(
+                argument -> {
+                  Value value = argument.getValue();
+                  return  argument.getName().equals("url")
+                      && value instanceof StringValue
+                      && ((StringValue)value).getValue().equals("https://specs.apollo.dev/federation/v2.0");
+                }
+            )
+        )
+    );
   }
 }
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -201,15 +201,11 @@ public final class FederationDirectives {
                 externalDefinition,
                 requiresDefinition,
                 providesDefinition,
-                extendsDefinition
-            )
+                extendsDefinition)
             .sorted(Comparator.comparing(DirectiveDefinition::getName))
             .collect(Collectors.toCollection(LinkedHashSet::new));
     federation2Definitions =
-        Stream.of(
-            shareableDefinition,
-            linkDefinition
-        )
+        Stream.of(shareableDefinition, linkDefinition)
             .sorted(Comparator.comparing(DirectiveDefinition::getName))
             .collect(Collectors.toCollection(LinkedHashSet::new));
     allNames =
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -186,6 +186,7 @@ public final class FederationDirectives {
   public static final Set<String> allNames;
   public static final Set<GraphQLDirective> allDirectives;
   public static final Set<DirectiveDefinition> allDefinitions;
+  public static final Set<DirectiveDefinition> federation2Definitions;
 
   static {
     // We need to maintain sorted order here for tests, since SchemaPrinter doesn't sort
@@ -200,9 +201,15 @@ public final class FederationDirectives {
                 externalDefinition,
                 requiresDefinition,
                 providesDefinition,
-                extendsDefinition,
-                shareableDefinition,
-                linkDefinition)
+                extendsDefinition
+            )
+            .sorted(Comparator.comparing(DirectiveDefinition::getName))
+            .collect(Collectors.toCollection(LinkedHashSet::new));
+    federation2Definitions =
+        Stream.of(
+            shareableDefinition,
+            linkDefinition
+        )
             .sorted(Comparator.comparing(DirectiveDefinition::getName))
             .collect(Collectors.toCollection(LinkedHashSet::new));
     allNames =
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
@@ -147,9 +147,6 @@ public final class FederationDirectives {
 
   public static final String shareableName = "shareable";
 
-  public static final GraphQLDirective shareable =
-      newDirective().name(shareableName).validLocations(FIELD_DEFINITION, OBJECT).build();
-
   public static final DirectiveDefinition shareableDefinition =
       newDirectiveDefinition()
           .name(shareableName)
@@ -164,9 +161,6 @@ public final class FederationDirectives {
    * refered to using `@federation__key`). """ import: [link__Import], ) repeatable on SCHEMA
    */
   public static final String linkName = "link";
-
-  public static final GraphQLDirective link =
-      newDirective().name(linkName).validLocations(SCHEMA).build();
 
   public static final DirectiveDefinition linkDefinition =
       newDirectiveDefinition()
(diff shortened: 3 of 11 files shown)


BoD (approved):
👍

sachindshinde (changes requested):

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -147,6 +143,48 @@ public static GraphQLDirective provides(String fields) {
...
+  /**
+   * directive @link( """ Url of the linked core feature. """ url: String!, """ Optional list of
+   * element names to import in the top-level namesapce. Elements of the feature not part of this
+   * list can only be referenced by prefixing the element named by the feature name (for instance,
+   * if the `@key` is not imported in this list for the `federation` feature, it can still be
+   * refered to using `@federation__key`). """ import: [link__Import], ) repeatable on SCHEMA
+   */
+  public static final String linkName = "link";
+
+  public static final GraphQLDirective link =
```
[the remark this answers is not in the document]
martinbonnin: Ah thanks!

I was wondering what all these `GraphQLDirective` were for. I'm still confused though... Do a lot of users define their schema programmatically? (Which is the main use case for the convenience functions  from what I understand?) That's sounds like a super verbose way to build your schema. Or are there other use cases?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
```diff
@@ -165,6 +165,7 @@ public static String sdl(GraphQLSchema schema, boolean queryTypeShouldBeEmpty) {
     hiddenTypeDefinitions.add(_Any.typeName);
     hiddenTypeDefinitions.add(_Entity.typeName);
     hiddenTypeDefinitions.add(_FieldSet.typeName);
+    hiddenTypeDefinitions.add(link__Import.typeName);
```
sachindshinde, line 168: So to give some history here, the reason why we remove directive and type definitions here is because very old versions of composition would fail if you provided those directive and type definitions in subgraph schemas.

This is because old versions of composition would unconditionally add those directive and type definitions, leading to duplicate definitions. Later composition versions fixed this.

Those old versions of composition would presumably fail if you removed the scalar type definition for `link__Import`, but kept the directive definition for `@link` (because they don't know anything about `@link` or `@shareable`, so they wouldn't add anything). I'm guessing that newer Fed 1 composition would also fail, so you may not want to remove anything related to `@link` and `@shareable` here.

We should also consider whether it would be fine to just drop support for such old versions of composition (and accordingly, such old versions of gateway that would run such old versions of composition). But that could be left to a follow-up PR if it's contentious. (The large benefit of that would be killing our schema printer fork, which exists solely because of type/directive definition filtering.)

sachindshinde (changes requested):

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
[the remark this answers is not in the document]
sachindshinde: >Do a lot of users define their schema programmatically?

I believe it's useful for libraries or places where enough people work on the codebase such that it becomes modularized.

Basically in such cases, you're often handed a `TypeDefinitionRegistry` or `GraphQLSchema` (or schema elements thereof), and you can't necessarily just ask the creator of those objects to modify their inputs or processing steps.

In fairness, I don't actually have real user feedback here though.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
martinbonnin: 👍 
Something we discussed with @pcmanus was to only add the Fed2 directives if the subgraph schema is targeting Fed2. From what I understand, this can be detecting by the presence of a schema extension:

```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@provides", "@external"])
```

([.ts tests here](https://github.com/apollographql/federation/blob/ac89f5fca2897757975cc545d3ea28f2a0919771/composition-js/src/__tests__/composeFed1Subgraphs.test.ts#L414))

I didn't do it in that PR but if it causes compatibility issues with Fed1, I might as well do it.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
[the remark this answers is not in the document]
sachindshinde: I'd be fine with that.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
sachindshinde: Conditionally adding the directives if the subgraph schema targets Fed 2 sounds fine to me.

Regarding the semantics, is it just the schema extension that triggers it? I imagine the condition for determining whether a subgraph is Fed 2 would be, "it has a schema or schema extension with an `@link` with `url` of `https://specs.apollo.dev/federation/v2.0`".

martinbonnin: > I imagine the condition for determining whether a subgraph is Fed 2 would be, "it has a schema or schema extension with an @link with url of https://specs.apollo.dev/federation/v2.0".

Yup, I'm speaking under @pcmanus supervision but this is what I understood. `extend schema` is just a convenient way to do it without having to specify the root query type.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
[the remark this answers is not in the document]
martinbonnin: The convenience have been removed, I'll re-add them if requested

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -168,7 +206,9 @@ private FederationDirectives() {}
                 externalDefinition,
                 requiresDefinition,
                 providesDefinition,
-                extendsDefinition)
+                extendsDefinition,
```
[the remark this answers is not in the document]
martinbonnin: Since I removed the `GraphQLDirective` for now (see [above](https://github.com/apollographql/federation-jvm/pull/144#discussion_r830073812)). I guess it's ok to not include them in `allDirectives`?

More generally, if there's a use case for `allDirectives` we should certainly add a matching test.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
martinbonnin: > Those old versions of composition would presumably fail if you removed the scalar type definition for link__Import, but kept the directive definition for @link

I think this is fine because the definition for `@link` is also removed. My understanding is that `_service.sdl` should print the subgraph schema omitting the federation directives definitions but including their usages.

> The large benefit of that would be killing our schema printer fork, which exists solely because of type/directive definition filtering.

I don't think we can do this even if we drop older versions? Mainly because of the above omission of the federation directives in the SDL. Looks like [subgraph-js has the same issue](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/buildSubgraphSchema.ts#L98).

> Only add the Fed2 directives if the subgraph schema is targeting Fed2. 

I checked [subgraph-js](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/buildSubgraphSchema.ts#L89) and the directives seem to be added unconditionally so I think it's better to just keep matching behaviour for now?

One thing that remains to be checked is the handling of schema extensions. Right now, something like

```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@provides", "@external"])
```

when read from `_service.sdl` is turned into 

```graphql
schema @link(import : ["@key", "@provides", "@external"], url : "https://specs.apollo.dev/federation/v2.0"){
  query: Query
}

type Query {
  _service: _Service
}
```

Not sure how much of an issue that is. I'll keep looking...

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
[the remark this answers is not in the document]
sachindshinde: @martinbonnin 
> Since I removed the `GraphQLDirective` for now (see [above](https://github.com/apollographql/federation-jvm/pull/144#discussion_r830073812)). I guess it's ok to not include them in `allDirectives`?

Yep, that sounds fine to me.

> More generally, if there's a use case for `allDirectives` we should certainly add a matching test.

I'm not sure of the use case, it was in the [initial commit for the repo](https://github.com/apollographql/federation-jvm/commit/f5a13182f76106d39d2674422abee8c599eea5a6#diff-d265303e57dcf25a6c597dde777163e1638e7bf140974a136afe87b8a300d36bR92) but it was unused then. I'm guessing it was meant to be a `GraphQLSchema`-equivalent for `allDefinitions`.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
martinbonnin: Reading more of [the spec](https://www.apollographql.com/docs/federation/federation-spec/), it says:

```
The generated SDL should match that exactly with no additions. It is important to preserve the type 
extensions and directive locations and to omit the federation types.
``` 

So it looks like this is important but doing some tests in https://github.com/apollographql/supergraph-demo-fed2 shows that composition still works with the current implementation.

This part is intringing: 

```
Some libraries such as graphql-java don't have native support for type extensions in their printer. 
Apollo Federation supports using an @extends directive in place of extend type to annotate type references:
```

That makes me wonder how much the 1:1 mapping from the user schema to the `sdl` field is important.

If it is, I guess we could make it work at the AST level (i.e. before we get a `GraphQLSchema` that loses the type extensions). All  in all, I'd keep this PR about directives and and refine schema printing in a separate one if needed.

pcmanus: > That makes me wonder how much the 1:1 mapping from the user schema to the `sdl` field is important.

Yeah, I left some comments on your PR for the spec as well, but as far as composition goes, it's not really that important. I think there is some legacy in there and I'm unfortunately not 100% sure how much of that is still necessary, but it would be nice to avoid making subgraph servers life harder than it needs to be.

sachindshinde: @martinbonnin 
> I think this is fine because the definition for `@link` is also removed.

Ah you're right, the definition for `@link` is also removed.

It's worth noting though that if we do that, we should also make sure that if
```graphql
schema @link(url: "https://specs.apollo.dev/federation/v1.0")
```
is given, we do one of the following to not break Fed 1 composition:
- Remove that `@link` directive application (since as I understand, Fed 2 composition assumes that subgraph schemas without `@link` are Fed 1, and Fed 1 composition never used/knew about `@link` anyway).
- Keep the definition for `@link` (and `link__Import`, if it's supposed to work for Fed 1).
- Error out, banning `"https://specs.apollo.dev/federation/v1.0"` from being used as the URL and saying it's unsupported for this library.

Not sure what's preferrable here (cc @pcmanus).

>My understanding is that `_service.sdl` should print the subgraph schema omitting the federation directives definitions but including their usages.

Sort of. There's some context here that's relevant:
1. For old gateways, this was because their composition would unconditionally add federation definitions, so including them would result in an error.
    a. Note this applies for just the federation directives that those gateways knew about. Other directives (federation or not) still needed definitions.
2. For current Fed 1 gateways, their composition should succeed regardless of whether you include federation definitions.
    a. There's an exception for `@tag`, where I think a directive definition is required.
3. For Fed 2 gateways, from what I understand you don't need the federation definitions. (I'm guessing composition won't fail if you include definitions though.)

> I don't think we can do this even if we drop older versions? Mainly because of the above omission of the federation directives in the SDL. Looks like [subgraph-js has the same issue](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/buildSubgraphSchema.ts#L98).

There's two possible desires here that I think are related but somewhat separate:
1. Users should be able to use federation directives without declaring any necessary federation type/directive definitions.
2. The subgraph schema at `_service.sdl` should be the same as that handed to the subgraph library.

The first desire I believe is less controversial, and it can be supported by this library without too much issue. We don't need a schema printer fork for that one.

The second one we don't currently honor today in `federation-jvm`. If we wanted to honor that one, we'd need the schema printer fork. It's also kind of hard to support (some programmatic representations of schemas in `graphql-java` don't allow the kinds of GraphQL-spec invalidities that subgraph schemas support).

I'm not sure how much we want to support the second desire (cc @pcmanus).

>I checked [subgraph-js](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/buildSubgraphSchema.ts#L89) and the directives seem to be added unconditionally so I think it's better to just keep matching behaviour for now?

That would be a breaking change for this library then, if e.g. the user had already declared user-defined directives `@link` and `@shareable` in Fed 1.

It could also lead to some confusing bugs. E.g. suppose someone on Fed 1 upgrades this library, and then they notice `@shareable` and `@link` definitions start appearing in their subgraph schemas. They skim through a blogpost and believe they can start using those directives, but they end up having no effect, which leaves them confused.

Maybe bugs using `@shareable` and `@link` aren't so bad, but believing e.g. subgraph `@inaccessible` works when it doesn't would be a rather bad security bug.

We could alternatively say, "this library only supports Fed 2 composition and gateways/routers," but not sure if we wanna go that route.

sachindshinde: @martinbonnin 


> One thing that remains to be checked is the handling of schema extensions. Right now, something like
> 
> ```graphql
> extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@provides", "@external"])
> ```
> 
> when read from `_service.sdl` is turned into
> 
> ```graphql
> schema @link(import : ["@key", "@provides", "@external"], url : "https://specs.apollo.dev/federation/v2.0"){
>   query: Query
> }
> 
> type Query {
>   _service: _Service
> }
> ```
> 
> Not sure how much of an issue that is. I'll keep looking...

I'm hoping this isn't too much of an issue (we've generally moved away from requiring anything have `extends`, since it has a specific meaning in the GraphQL spec, and co-opting it to also mean something special in federation has proved painful).

martinbonnin:
```graphql
schema @link(url: "https://specs.apollo.dev/federation/v1.0")
```

Is that even possible? That'd be using a fed2 directive to tell a gateway to use fed1 composition? I'm ok throwing an error on this.

> The subgraph schema at _service.sdl should be the same as that handed to the subgraph library.

That's the hard part. The definition of "same" is hard to get right. 

* Is it ok to merge type extensions?
* Is it ok to add a schema definition if none is present? 
* Is it ok to add `_Entity`, `Query._entities`, `@key` and others?  

@pcmanus [pointed out](https://github.com/apollographql/federation/pull/1627#discussion_r830987242) that fed2 might very well support subgraphs printing all of that in `_service.sdl` which would mean that it becomes "valid" SDL again and we _could_ (maybe) use more upstream printers 🤞

> Maybe bugs using @shareable and @link aren't so bad, but believing e.g. subgraph @inaccessible works when it doesn't would be a rather bad security bug.

Wait, what is `@inaccessible` 😅  ? It's not added by [subgraph-js](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/directives.ts#L105) so I didn't add it in this PR.

Overall, I think consistency is key (no pun intended) and subgraph-js being the de-facto implementation I'd say as long as we stick to that we should be good? Or is the landscape that different on the JVM?

sachindshinde: @martinbonnin 
> Reading more of [the spec](https://www.apollographql.com/docs/federation/federation-spec/), it says:
> 
> ```
> The generated SDL should match that exactly with no additions. It is important to preserve the type 
> extensions and directive locations and to omit the federation types.
> ```
> 
> So it looks like this is important but doing some tests in https://github.com/apollographql/supergraph-demo-fed2 shows that composition still works with the current implementation.

The subgraph spec is unfortunately a bit old there. There's three specific asks:
1. "It is important to preserve the type extensions"
    - I believe this is more about not removing `extend`, which we eventually relaxed because `extend` has special meaning in the GraphQL spec and no libraries preserve it. With `federation-jvm`, we instead support the directive `@extends`.
2. "and directive locations"
    - I believe this is about not removing any directive applications from the subgraph SDL (some GraphQL libraries don't print directive applications in some cases, e.g. graphql-js).
3. "and to omit the federation types."
    - I believe this is back when composition would unconditionally add them, so it was a necessity then.

More generally, the spec hasn't been appropriately versioned and documented to account for compatibility concerns, so it's sort of this vague thing in the minds of subgraph library developers at the moment.

> This part is intringing:
> 
> ```
> Some libraries such as graphql-java don't have native support for type extensions in their printer. 
> Apollo Federation supports using an @extends directive in place of extend type to annotate type references:
> ```
> 
> That makes me wonder how much the 1:1 mapping from the user schema to the `sdl` field is important. (edit: [link to the spec PR](https://github.com/apollographql/federation/pull/1627))

Yeah, trying to print the user's pre-transform schema 1:1 would be difficult/require a fair amount of work.

> If it is, I guess we could make it work at the AST level (i.e. before we get a `GraphQLSchema` that loses the type extensions).

We could, although in some cases users may have already transformed their schema into a `GraphQLSchema` due to either their own code/schema build process, or due to other libraries that work with `GraphQLSchema`s instead of ASTs (e.g. this library outputs a `GraphQLSchema` instead of an AST).

> All in all, I'd keep this PR about directives and and refine schema printing in a separate one if needed.

Yeah, no need to worry about the printer in this PR.

> ```graphql
> schema @link(url: "https://specs.apollo.dev/federation/v1.0")
> ```
> Is that even possible? That'd be using a fed2 directive to tell a gateway to use fed1 composition? I'm ok throwing an error on this.

To clarify, I think using `@link(url: "https://specs.apollo.dev/federation/v1.0")` would be more about specifying the versions of federation subgraph directives/behavior. It doesn't directly dictate composition version; people can mix and match subgraphs that use `"https://specs.apollo.dev/federation/v1.0"` and  `"https://specs.apollo.dev/federation/v2.0"` in Fed 2 composition if I understand correctly. But a subgraph using `"https://specs.apollo.dev/federation/v2.0"` requires Fed 2 composition.

In any case, I'd say erroring on values that aren't `"https://specs.apollo.dev/federation/v2.0"` would be fine.

> @pcmanus [pointed out](https://github.com/apollographql/federation/pull/1627#discussion_r830987242) that fed2 might very well support subgraphs printing all of that in `_service.sdl` which would mean that it becomes "valid" SDL again and we _could_ (maybe) use more upstream printers 🤞

That would definitely be nice.

That would also resolve the GraphQL-invalidity issue where pre-transform subgraph schemas are allowed to omit a `Query` type (or leave it empty), which is invalid GraphQL, but post-transform subgraph schemas have `Query._service` and don't have this issue.

There's again compatibility concerns I believe (e.g. does Fed 1 composition play nicely with fully-printed subgraph SDLs?), so we'd need to figure out how much support we want to offer there.

> Wait, what is `@inaccessible` 😅 ? It's not added by [subgraph-js](https://github.com/apollographql/federation/blob/18468bb3322c0b7c636f9a11f5cc0702e79b2f65/subgraph-js/src/directives.ts#L105) so I didn't add it in this PR.

No need to worry about that for now, it's a directive that some Atlas folk are in the midst of supporting for subgraphs for Fed 2 GA.

>Overall, I think consistency is key (no pun intended) and subgraph-js being the de-facto implementation I'd say as long as we stick to that we should be good? Or is the landscape that different on the JVM?

In a world where this library was brand new, I'd definitely be down to just copy subgraph-js. It's harder now though since this library has a decent number of users (some direct, and some indirect via DGS and graphql-kotlin).

I'm not saying we can't pivot toward being consistent with subgraph-js, but we'd have to account for:
- Differences in graphql-js and graphql-java that may lead to inconsistencies.
- Differences in support guarantees, particularly in places where we support old use cases that aren't supported by subgraph-js. We may be fine dropping support there, but we should make sure to notify folks in the changelog (and potentially reach out to customers/users where appropriate).

pcmanus: > In a world where this library was brand new, I'd definitely be down to just copy subgraph-js. It's harder now though since this library has a decent number of users

As a warning, I'd also be careful with holding current `subgraph-js` implementation as a model to follow too closely. In fact, the plan of record is to have a lot of it rewritten (by https://github.com/apollographql/federation/pull/1554) before GA for fed2 which may/may not change some of the details discussed here.

> Differences in support guarantees, particularly in places where we support old use cases that aren't supported by subgraph-js.

I think that's what @sachindshinde is eluding to, but quick reminder that `subgraph-js` is essentially forked into a `0.x` and a `2.x` version, the former still being independently supported. That versioning makes it a bit more ok for the `2.x` version to not worry too closely about changing some of the outputs, but if this library don't want to bother with such forking, the context might be a tad different.

martinbonnin: > the plan of record is to have a lot of it rewritten (by https://github.com/apollographql/federation/pull/1554) before GA for fed2 which may/may not change some of the details discussed here

Sounds good 👍 . Then maybe we should work together on this and refine the implementation details as we go to make sure we ship matching implementations? 

> if this library don't want to bother with such forking, the context might be a tad different.

No strong opinion on this one. Feels like having backward compatibility is nice but if the maintainance cost is too high we could as well fork.

All in all, I'd rather wait a bit a ship something that we can trust and build upon rather than ship something now that we will regret down the road.

graphql-java-support/pom.xml
```diff
@@ -70,7 +70,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <protocArtifact>com.google.protobuf:protoc:3.9.0:exe:${os.detected.classifier}</protocArtifact>
+                    <protocArtifact>com.google.protobuf:protoc:3.19.4:exe:${os.detected.classifier}</protocArtifact>
```
martinbonnin, line 73: This is not related to Fed2. It's mainly to keep the `protoc` version in sync with the runtime version above and also fixes compiling on Mac M1s

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
```diff
@@ -43,12 +41,14 @@ public static SchemaTransformer transform(
...
+
+    boolean isFederation2 = isFederation2(typeRegistry);
     RuntimeWiring newRuntimeWiring =
-        ensureFederationDirectiveDefinitionsExist(typeRegistry, runtimeWiring);
+        ensureFederationDirectiveDefinitionsExist(typeRegistry, runtimeWiring, isFederation2);
     final GraphQLSchema original =
         new SchemaGenerator()
             .makeExecutableSchema(generatorOptions, typeRegistry, newRuntimeWiring);
-    return transform(original, queryTypeShouldBeEmpty);
+    return transform(original, queryTypeShouldBeEmpty).setFederation2(isFederation2);
```
martinbonnin, line 51: Because we get `isFederation` from the `typeRegistry`, this does not apply to `Federation.transform(GraphQLSchema)`. I'm not sure who's using that though. DGS uses [`Federation.transform(typeRegistry)`](https://github.com/netflix/dgs-framework/blob/595f2a8762a53f321c9285b331bf9196958e2cc8/graphql-dgs/src/main/kotlin/com/netflix/graphql/dgs/internal/DgsSchemaProvider.kt#L134) and graphql-kotlin [uses its own implementation](https://github.com/ExpediaGroup/graphql-kotlin/blob/de02352dc8af5bc6bb78b90d8d8f4908988999f6/generator/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/generator/federation/FederatedSchemaGeneratorHooks.kt#L82).

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
```diff
@@ -184,4 +201,25 @@ private static RuntimeWiring ensureFederationDirectiveDefinitionsExist(
...
+                                    .anyMatch(
+                                        argument -> {
+                                          Value value = argument.getValue();
+                                          return argument.getName().equals("url")
+                                              && value instanceof StringValue
+                                              && ((StringValue) value)
+                                                  .getValue()
+                                                  .equals(
+                                                      "https://specs.apollo.dev/federation/v2.0");
+                                        })));
```
martinbonnin, line 223: Yikes that's a nice pyramid of doom 🙈 . If there's a way to improve this, I'll take it!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -1,20 +1,14 @@
...
 import static graphql.schema.GraphQLArgument.newArgument;
 import static graphql.schema.GraphQLDirective.newDirective;
 
 import graphql.PublicApi;
-import graphql.language.DirectiveDefinition;
-import graphql.language.DirectiveLocation;
-import graphql.language.InputValueDefinition;
-import graphql.language.NonNullType;
-import graphql.language.TypeName;
+import graphql.language.*;
```
martinbonnin, line 11: My intelliJ did that. Are we OK with star imports? If not we should certainly commit some project configuration somewhere so we have all the same.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
martinbonnin: I just pushed a few modifications to:

1. Detect the Fed2 header so that modifications are only done for Fed2 schemas. The Fed1 path should be unchanged so that we keep 100% backward compat.
2. Add the `@tag` directive
3. Include the federation directives in `_service.sdl` according to [this change in subgraph-js](https://github.com/apollographql/federation/pull/1554). This way the document in `_service.sdl` is valid SDL

Next steps:

1. I'd like to merge this PR and release a `2.0.0-alpha.1` so that Fed2 support can be easily tested and we can get feedback. 
2. Get feedback and see what's missing? Maybe `@inaccessible`, something else? 
3. Support for renaming the directives and better core schema suport. Maybe write `core-schema-jvm` like there is https://github.com/apollographql/core-schema-js ?

Let me know what you think.

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -147,13 +145,64 @@ public static GraphQLDirective provides(String fields) {
...
+                  .type(new ListType(new TypeName("link__Import")))
+                  .build())
+          .repeatable(true)
+          .build();
+
+  /** directive @tag(name: String!) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION */
+
+  /**
+   * + * directive @tag(name: String!) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION +
+   */
```
BoD, line 188: Typo?
```suggestion
  /** directive @tag(name: String!) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION */
```

martinbonnin: Yep, bad copy/paste, thanks!

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
BoD: To be fair I think the "pyramid" effect is due to the formatter 😁. You could maybe extract a few bits in dedicated methods but to me it's readable enough as-is.

martinbonnin: Yes, exactly! I had it with a lot less levels of indentation before I ran `./mvnw spotless:apply` 😬

BoD (approved):

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
hwillson: @martinbonnin please proceed with merging / getting the `2.0.0-alpha.1` release ready. Thanks!

PR #147, opened 2022-03-29
Use bom version 2.13.2.20220328 for Jackson dependencies

BoD:
Following [this Dependabot PR](https://github.com/apollographql/federation-jvm/pull/146) which tries to address [this CVE](https://github.com/FasterXML/jackson-databind/issues/2816) but bumps the dependencies globally which doesn't work - this PR uses [the BOM](https://github.com/FasterXML/jackson-bom) instead.

### DIFF
--- a/spring-example/pom.xml
+++ b/spring-example/pom.xml
@@ -72,28 +72,6 @@
                 <scope>test</scope>
             </dependency>
             <!--
-            The following dependencies exist solely to override Maven's default "nearest first" strategy for transitive
-            dependency version conflict resolution. Note that as stated in the parent POM, we would normally use Maven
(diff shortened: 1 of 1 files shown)


spring-example/pom.xml
```diff
@@ -65,32 +72,6 @@
                 <scope>test</scope>
             </dependency>
             <!--
-            The following dependencies exist solely to override Maven's default "nearest first" strategy for transitive
```
BoD, line 68: I must admit I am not sure why these were needed, but it seems to work without them.

dariuszkuc:
Superseded by https://github.com/apollographql/federation-jvm/pull/195.

Issue #165, opened 2022-04-25
Support link spec import aliasing


hwillson:
[Version 1.0 of the `link` spec](https://github.com/apollographql/specs/blob/collect-specs/link/v1.0/link-v1.0.md) (used by Federation 2) supports [directive renaming](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#renaming-directives) using `as`:

```graphql
@link(url: "https://example.com/, import: [{
  name: "SomeType",
  as: "@someDirective"
}])
```

Federation JVM will need to support this as well.

PR #166, opened 2022-04-29
Add support for rename imports

martinbonnin:
This adds support for renaming directives like:

```graphql
extend schema
    @link(url: "https://specs.apollo.dev/federation/v2.0",
       import: ["@key", "@shareable", {name: "@tag" as :"@apollo_tag"}, "@override"])
```

If the directive is not imported, it can still be used with the `federation__` prefix:

```graphql
type Position @federation__external
```

Fixes https://github.com/apollographql/federation-jvm/issues/165

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
@@ -222,10 +222,10 @@ public final class Federation {
           if (def instanceof ScalarTypeDefinition
               && !runtimeWiring.getScalars().containsKey(def.getName())) {
             scalarTypesToAdd.add(
-                GraphQLScalarType.newScalar(Scalars.GraphQLString)
+                GraphQLScalarType.newScalar()
                     .name(def.getName())
                     .description(null)
-                    .coercing(Scalars.GraphQLString.getCoercing())
+                    .coercing(_Any.defaultCoercing)
                     .build());
           }
         });
@@ -394,7 +394,7 @@ public final class Federation {
                         as = null;
                       } else {
                         Value asValue = asField.get().getValue();
-                        if (asValue instanceof StringValue) {
+                        if (!(asValue instanceof StringValue)) {
                           throw new RuntimeException("Unsupported import: " + imp);
                         }
                         as = ((StringValue) asValue).getValue();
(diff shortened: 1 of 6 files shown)


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
```diff
@@ -132,14 +137,106 @@ private static boolean ensureQueryTypeExists(TypeDefinitionRegistry typeRegistry
...
+
+  private static String newName(String name, Map<String, String> fed2Imports, boolean isDirective) {
+    String key;
+    if (isDirective) {
+      key = "@" + name;
+    } else {
+      key = name;
+    }
+
+    if (key.equals("String") || key.equals("Boolean") || key.equals("Int") || key.equals("Float")) {
```
BoD, line 160: Should `"ID"` be there too?

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
```diff
@@ -171,98 +143,36 @@ public static GraphQLDirective provides(String fields) {
...
   public static final Set<String> allNames;
   public static final Set<GraphQLDirective> allDirectives;
   public static final Set<DirectiveDefinition> allDefinitions;
-  public static final Set<DirectiveDefinition> federation2DirectiveDefinitions;
+  static final List<SDLNamedDefinition> federation2Definitions;
   public static final Set<DirectiveDefinition> federation1DirectiveDefinitions;
 
+  private static List<SDLNamedDefinition> fed2Definitions() {
+    InputStream inputStream =
+        FederationDirectives.class.getClassLoader().getResourceAsStream("fed2directives.graphqls");
```
BoD, line 158: Very cool, a lot simpler / less verbose!

BoD (approved):
🚀

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/Federation.java
martinbonnin: uh yes, good catch 👍

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationDirectives.java
martinbonnin: Pretty happy about that part :)

PR #177, opened 2022-05-25
[build] migrate to Gradle

dariuszkuc:
Resolves https://github.com/apollographql/federation-jvm/issues/131


martinbonnin (approved):
I'm off for a long week end so just made a very quick pass at this but I don't want to block it, LGTM!  Super happy this is happening 🐘 🚀

gradle.properties
```diff
@@ -0,0 +1,22 @@
+group = com.apollographql.federation
+version = 2.0.2-SNAPSHOT
+
+# See https://github.com/gradle/gradle/issues/11308
+systemProp.org.gradle.internal.publish.checksums.insecure=true
```
martinbonnin, line 5: I think you can skip this. Sonatype now handles this correctly

Issue #169, opened 2022-05-04
Spotless fails on Java 16+ 


martinbonnin:
Follow up from https://github.com/diffplug/spotless/issues/834. 

There's a lot of information in https://github.com/apollographql/federation-jvm/issues/167 about a possible workaround.

dariuszkuc:
Resolved in https://github.com/apollographql/federation-jvm/pull/209

dariuszkuc:
Still an issue.

dariuszkuc:
Resolved in https://github.com/apollographql/federation-jvm/pull/210

PR #207, opened 2022-06-03
[build] add JVM flags to allow running spotless in JDK 16+

dariuszkuc:
Resolves: https://github.com/apollographql/federation-jvm/issues/169


dariuszkuc:
Those flags are not available on JDK8 (which we currently use to build the lib). We could add this fix if we decide to build on newer JVMs. Closing for now.

PR #212, opened 2022-06-14
[build] fix intellij configuration

dariuszkuc:
Protobuf plugin does not correctly register its generated sources meaning that Intellij does not see the generated classes. Updating configuration to explicitly register generated proto sources.


martinbonnin:
Just curious if you have any thoughts on [wire](https://square.github.io/wire/) from Square? IIRC they do this automagically.

PR #223, opened 2022-07-11
use GraphQLAppliedDirective instead of GraphQLDirective for printing FederationSDL

nagliyvred:
Latest graphql-java is reimplementing directives https://github.com/graphql-java/graphql-java/pull/2562 and moving to GraphQLAppliedDirective instead. 
Similarly, graphql-java-kickstart project is being updated to make use of GraphQLAppliedDirective too https://github.com/graphql-java-kickstart/graphql-java-tools/pull/663 . 
This breaks federation, as the schema built with the latest graphql-java-tools would not contain any GraphQLDirective and therefore generated SDL is going to have all of the federation-related directives missing. 
This PR is meant to address it by getting FedarationSdlPrinter to use the new directive classes everywhere, except where the actual directive definition is printed.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/FederationSdlPrinter.java
@@ -2,10 +2,6 @@ package com.apollographql.federation.graphqljava;
 
 import static graphql.Directives.DeprecatedDirective;
 import static graphql.com.google.common.base.Predicates.not;
-import static graphql.introspection.Introspection.DirectiveLocation.ARGUMENT_DEFINITION;
-import static graphql.introspection.Introspection.DirectiveLocation.ENUM_VALUE;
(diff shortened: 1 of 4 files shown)


dariuszkuc:
👋 thanks for raising this, indeed it looks like I missed it when I fixed the entity lookup logic in https://github.com/apollographql/federation-jvm/pull/214. 

That SDL printer is a copy from `graphql-java` v17.3, since we are using `graphql-java` v18 now it would be great to update it to the latest.

Two considerations
* for federation v2 we won't need custom printer once `graphql-java` v19 is released (there is a bug in v18 that does not print schema directives)
* for federation v1 we need custom logic to skip federation directive definitions

Would you be willing to update the printer to the latest https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/SchemaPrinter.java?

dariuszkuc:
Superseded by https://github.com/apollographql/federation-jvm/pull/231

PR #246, opened 2022-09-19
fix: allow @link on schema type

dariuszkuc:
Fixed bugs with how we process `@link` directives.

`@link` directive can be applied directly on schema type. Previous logic only looked for the `@link` directive on schema extension.

When we are importing @link definitions we can use either String or an Object that allows the rename. Existing logic was throwing NPE when import object was specified without renamed value (see # 3 below).

1. using imports directly -> `@link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key"])`
2. using object renames -> `@link(url: "https://specs.apollo.dev/federation/v2.0", import: [{ name: "@key", as: "@myKey" }])`
3. using object without rename -> `@link(url: "https://specs.apollo.dev/federation/v2.0", import: [{ name: "@key" }])`


hwillson (approved):
Looks great to me @dariuszkuc - thanks!

PR #249, opened 2022-10-10
add support for federation v2.1

dariuszkuc:
Add support for latest Federation spec versions and improve validation logic. See [Federation documentation](https://www.apollographql.com/docs/federation/) for additional details.

Federation v2.1 features:

* support new `@composeDirective`

New validations:

* Multiple `@link` imports of federation spec are no longer valid.
* Attempting to import `@composeDirective` from Fed v2.0 spec will throw validation exception.
* Attempting to rename `@tag`/`@inaccessible` now throws more specific exception
* Validating federation spec version

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/directives/LinkImportsRenamingVisitor.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/directives/LinkImportsRenamingVisitor.java
@@ -1,7 +1,8 @@
-package com.apollographql.federation.graphqljava;
+package com.apollographql.federation.graphqljava.directives;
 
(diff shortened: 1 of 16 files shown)


graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/directives/LinkImportsRenamingVisitor.java
```diff
@@ -1,7 +1,8 @@
-package com.apollographql.federation.graphqljava;
+package com.apollographql.federation.graphqljava.directives;
```
dariuszkuc, line 1: renaming visitor has a default class visibility so it is not part of the public API and can be moved to different package without causing a breaking change

PR #285, opened 2023-01-06
feature: auto-populate entity type resolver for non-resolvable entities

### LINKED ISSUE #70: Question to the SchemaTransformer logic
Hello, just wonder if you can help me understand the logic behind the following requirement:
```
If your schema does not contain any types annotated with the @key directive, nothing else is required. You can build a transformed GraphQLSchema with SchemaTransformer#build(), and confirm it exposes query { _schema { sdl } }.

Otherwise, all types annotated with @key will be part of the _Entity union type, and reachable through query { _entities(representations: [Any!]!) { … } }
```

According to the [specification](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/), if I extend another type I have to add **@key** directive. But in case if I just want to reference another type and I don't extend it with any additional attributes, why do I have to define data fetcher that will never be invoked? 

Moreover, I've just looked through the code of SchemaTransformer.java, and it doesn't validate that particular type with directive @key defined has data fetcher, it is enough if there is at least one defined.

This looks like something that need to be changed.

dariuszkuc:
Whenever all entites are non-resolvable locally (i.e. `@key` specifies `resolvable:false`), then corresponding resolvers should never be invoked locally and users shouldn't have to specify fake ones.

Changes:
* default fake `_Entity` type resolver will be automatically created if it is not provided and all entities are non-resolvable
* default fake `_Entity` type resolver will throw `IllegalStateException` if invoked at runtime
* `_entities` data fetcher does not have to be specified if all entities are non-resolvable

Related:
* Resolves #70

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/SchemaTransformer.java
@@ -247,41 +247,41 @@ public final class SchemaTransformer {
 
   private boolean resolvableEntitiesExist(Set<String> entityNames) {
     return entityNames.stream()
-      .anyMatch(
-        entity -> {
-          GraphQLObjectType entityObject = (GraphQLObjectType) originalSchema.getType(entity);
-          boolean isResolvable =
-            entityObject.getAppliedDirectives(FederationDirectives.keyName).stream()
-              .anyMatch(
-                key -> {
-                  GraphQLAppliedDirectiveArgument resolvable =
-                    key.getArgument("resolvable");
-                  if (resolvable != null) {
-                    BooleanValue resolvableValue =
-                      (BooleanValue) resolvable.getArgumentValue().getValue();
-                    return resolvableValue == null || resolvableValue.isValue();
-                  } else {
-                    return true;
-                  }
-                });
+        .anyMatch(
+            entity -> {
+              GraphQLObjectType entityObject = (GraphQLObjectType) originalSchema.getType(entity);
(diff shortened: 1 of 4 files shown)


dariuszkuc:
## Apollo Federation Subgraph Compatibility Results

<table>
<thead>
<tr><th>Federation 1 Support</th><th>Federation 2 Support</th></tr>
</thead>
<tbody>
<tr><td><table><tr><th><code>_service</code></th><td>🟢</td></tr><tr><th><code>@key (single)</code></th><td>🟢</td></tr><tr><th><code>@key (multi)</code></th><td>🟢</td></tr><tr><th><code>@key (composite)</code></th><td>🟢</td></tr><tr><th><code>repeatable @key</code></th><td>🟢</td></tr><tr><th><code>@requires</code></th><td>🟢</td></tr><tr><th><code>@provides</code></th><td>🟢</td></tr><tr><th><code>federated tracing</code></th><td>🟢</td></tr></table></td><td><table><tr><th><code>@link</code></th><td>🟢</td></tr><tr><th><code>@shareable</code></th><td>🟢</td></tr><tr><th><code>@tag</code></th><td>🟢</td></tr><tr><th><code>@override</code></th><td>🟢</td></tr><tr><th><code>@inaccessible</code></th><td>🟢</td></tr></table></td></tr>
</tbody>
</table>

Learn more:
* [Apollo Federation Subgraph Specification](https://www.apollographql.com/docs/federation/subgraph-spec/)
* [Compatibility Tests](https://github.com/apollographql/apollo-federation-subgraph-compatibility/blob/main/COMPATIBILITY.md)

PR #293, opened 2023-02-15
fix: update PR check workflows to correctly check out codebase

dariuszkuc:


.github/workflows/build.yaml
```diff
@@ -14,10 +14,10 @@ jobs:
       - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1
 
-      - name: Set up Java 8
+      - name: Set up Java 17
```
dariuszkuc, line 17: we are building with Java 17 but we are still targeting Java 8 bytecode

PR #347, opened 2023-09-05
feat: federation 2.5 support

dariuszkuc:
Adds support for [Federation specification v2.5](https://www.apollographql.com/docs/federation/federation-versions#v25).

Changes:

* new `@authenticated` directive

```graphql
directive @authenticated on
    ENUM
  | FIELD_DEFINITION
  | INTERFACE
  | OBJECT
  | SCALAR
```

* new `@requiresScopes` directive

```graphql
directive @requiresScopes(scopes: [[Scope!]!]!) on
    ENUM
  | FIELD_DEFINITION
  | INTERFACE
  | OBJECT
  | SCALAR

scalar Scope
```

### DIFF
--- a/graphql-java-support/src/main/resources/definitions_fed2_5.graphqls
+++ b/graphql-java-support/src/main/resources/definitions_fed2_5.graphqls
@@ -1,0 +1,83 @@
+#
+# https://specs.apollo.dev/federation/v2.0/federation-v2.0.graphql
+#
+
+directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
+directive @requires(fields: FieldSet!) on FIELD_DEFINITION
+directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+directive @external on OBJECT | FIELD_DEFINITION
+directive @extends on OBJECT | INTERFACE
+directive @override(from: String!) on FIELD_DEFINITION
+directive @inaccessible on
+    | FIELD_DEFINITION
+    | OBJECT
+    | INTERFACE
+    | UNION
+    | ENUM
+    | ENUM_VALUE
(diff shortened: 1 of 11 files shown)


graphql-java-support/src/main/resources/definitions_fed2_5.graphqls
```diff
@@ -0,0 +1,83 @@
...
+  | SCALAR
+
+directive @requiresScopes(scopes: [[Scope!]!]!) on
+    ENUM
+  | FIELD_DEFINITION
+  | INTERFACE
+  | OBJECT
+  | SCALAR
+
+scalar Scope
```
trevor-scheer, line 83: Pretty sure this needs to be `federation__Scope` everywhere it's defined and referenced by `@requiresScopes`. I would expect a composition error / incompatible definition for this.

```
Invalid definition for directive "@requiresScopes": argument "scopes" should have type "[[federation__Scope!]!]!" but found type "[[Scope!]!]!"
```

dariuszkuc: `federation__` namespace prefix should only be required if given directive/type is not explicitly imported (see the diff between the [custom authenticated directive with no imports](https://github.com/apollographql/federation-jvm/blob/379008e354d21203da1a33a8d73c8e339a547035/graphql-java-support/src/test/resources/schemas/customAuthenticated_federated.graphql#L23) and [default authorization with imports](https://github.com/apollographql/federation-jvm/blob/379008e354d21203da1a33a8d73c8e339a547035/graphql-java-support/src/test/resources/schemas/authorization_federated.graphql#L29)

trevor-scheer (approved):
LGTM. I'm not really privy to the java-related changes but they seem reasonable.

PR #354, opened 2023-10-04
feat: subscription http callback support for spring graphql

dariuszkuc:
GraphQL subscriptions enable clients to receive continual, real-time updates whenever new data becomes available. Unlike queries and mutations, subscriptions are long-lasting. This means a client can receive multiple updates from a single subscription.

[Spring GraphQL](https://docs.spring.io/spring-graphql/reference/) provides out of box support for GraphQL subscriptions over WebSockets using [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws) protocol. 

This PR adds new `federation-spring-subscription-callback` module that provides support for subscriptions using [Apollo HTTP callback protocol](https://www.apollographql.com/docs/router/executing-operations/subscription-callback-protocol).

See
* [Apollo Router](https://www.apollographql.com/docs/router/executing-operations/subscription-support) for additional details about Federation and Subscription support


spring-subscription-callback/src/main/java/com/apollographql/subscription/callback/SubscriptionCallbackHandler.java
```diff
@@ -0,0 +1,201 @@
...
+        .bodyValue(checkMessage)
+        .exchangeToMono(
+            checkResponse -> {
+              var responseStatusCode = checkResponse.statusCode();
+              var subscriptionProtocol =
+                  checkResponse.headers().header(SUBSCRIPTION_PROTOCOL_HEADER);
+
+              if (responseStatusCode.is2xxSuccessful()) {
+                //            && !subscriptionProtocol.isEmpty() &&
+                // "callback".equals(subscriptionProtocol.get(0)))
```
dariuszkuc, line 70: router currently does not respond with this header

spring-subscription-callback/README.md
```diff
@@ -0,0 +1,132 @@
...
+queries and mutations, subscriptions are long-lasting. This means a client can receive multiple updates from a single subscription.
+
+[Spring GraphQL](https://docs.spring.io/spring-graphql/reference/) provides out of box support for GraphQL subscriptions
+over WebSockets using [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws) protocol. This library adds support
+for subscriptions using [Apollo HTTP callback protocol](https://www.apollographql.com/docs/router/executing-operations/subscription-callback-protocol).
+
+See [Apollo Router](https://www.apollographql.com/docs/router/executing-operations/subscription-support) for additional
+details about Federation and Subscription support.
+
+```mermaid
```
dariuszkuc, line 13: rendered -> https://github.com/dariuszkuc/federation-jvm/tree/callback/spring-subscription-callback

spring-subscription-callback/src/test/java/com/apollographql/subscription/callback/SubscriptionCallbackHandlerTest.java
```diff
@@ -0,0 +1,369 @@
...
+
+      var graphQLRequest = stubWebGraphQLRequest(subscriptionId, callbackUrl);
+      var data =
+          Flux.just(1, 2)
+              .delayElements(Duration.ofMillis(3000))
+              .map((i) -> ExecutionResult.newExecutionResult().data(Map.of("counter", i)).build());
+      var handler = new SubscriptionCallbackHandler(new MockWebHandler(data));
+
+      // note: heartbeat goes into infinite recursion and does not emit value
+      // TODO update to use virtual timer
```
dariuszkuc, line 190: unsure why but virtual timer didn't work... this can be fixed in the future (as it is right now those tests take couple seconds to run....)

spring-subscription-callback/src/main/java/com/apollographql/subscription/callback/SubscriptionCallbackHandler.java
```diff
@@ -0,0 +1,201 @@
...
+
+  public SubscriptionCallbackHandler(WebGraphQlHandler graphQlHandler, Scheduler scheduler) {
+    this.graphQlHandler = graphQlHandler;
+    this.scheduler = scheduler;
+  }
+
+  @NotNull
+  public Mono<Boolean> handleSubscriptionUsingCallback(
+      @NotNull WebGraphQlRequest graphQlRequest, @NotNull SubscriptionCallback callback) {
+    if (logger.isDebugEnabled()) {
```
clenfest, line 48: Do you really need to wrap this in `isDebugEnabled`? Shouldn't `logger.debug` do this check for you?

clenfest (approved):
I didn't really drill into the nitty gritty, but the logic looks good to me.

spring-subscription-callback/src/main/java/com/apollographql/subscription/callback/SubscriptionCallbackHandler.java
dariuszkuc: In general yes as otherwise you end up creating the log string regardless whether you will be logging it or not (I guess you could rewrite log message to use a string template to avoid this but in general it is a good practice so we don't forget it in the future).

PR #360, opened 2023-11-22
feat(callback): support configurable heartbeat interval

dariuszkuc:
Update protocol to support configurable heartbeat interval.


bnjjj:
I think that would be interesting to handle the case when `heartbeat_interval_ms` is not provided. To have a smooth transition for people using an older version of the router and stay backward compatible during few weeks. What do you think @dariuszkuc ?

bnjjj (approved):

PR #368, opened 2024-01-18
Default to not tracing

smyrick:
Federation tracing should only be enabled if the Router/Gateway sends the proper ftv1 header and value. We want to default to false so we are not wasting CPU cycles tracking resolvers and building the extensions if it not going to be used

https://www.apollographql.com/docs/federation/metrics/

We need tests. Just making this to start discussion


dariuszkuc:
Superseded by #370

PR #392, opened 2024-05-14
feat: add flag to CacheControlInstrumentation to optionally allow max-age of zero

trophied:
This change adds an optional flag to `CacheControlInstrumentation`, `allowZeroMaxAge`, to allow consumers to configure the behavior such that resolved header can validly resolve a header value of `max-age=0`.

### DIFF
--- a/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/caching/CacheControlInstrumentation.java
+++ b/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/caching/CacheControlInstrumentation.java
@@ -29,6 +29,7 @@ import org.jetbrains.annotations.Nullable;
  */
 public class CacheControlInstrumentation extends SimplePerformantInstrumentation {
   private final int defaultMaxAge;
+  private final boolean allowZeroMaxAge;
 
   private static final Object CONTEXT_KEY = new Object();
   private static final String DIRECTIVE_NAME = "cacheControl";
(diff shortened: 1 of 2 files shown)


dariuszkuc (approved):

dariuszkuc:
Linter failed -> can you run `./gradlew spotlessApply` (this should apply auto format, there might be other linter issues that need manual fixing).

trophied:
> Linter failed -> can you run `./gradlew spotlessApply` (this should apply auto format, there might be other linter issues that need manual fixing).

Sure thing, I must have missed re-running after changing some stuff out 👍

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/caching/CacheControlInstrumentation.java
```diff
@@ -223,11 +238,13 @@ void replace(@Nullable CacheControlScope scope) {
 
     public Optional<String> maybeAsString() {
       Integer maxAgeValue = maxAge == null ? 0 : maxAge;
-      if (maxAgeValue.equals(0)) {
+
+      if (!allowZeroMaxAge && maxAgeValue.equals(0)) {
```
dariuszkuc, line 242: actually I think we should change the logic here -> we should detect between `maxAge` is `null` and `0` so I think we probably should have the condition as

```java
if (maxAge == null || (!allowZeroMaxAge && maxAge == 0)) {
  return Optional.empty();
}
```

dariuszkuc (changes requested):

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/caching/CacheControlInstrumentation.java
trophied: Sounds good, I have added the explicit handling of `null` 👍

dariuszkuc (approved):

PR #395, opened 2024-05-16
breaking: update to graphql-java 22

dariuszkuc:
The only breaking change is to the caching instrumentation that relied on deprecated methods.

Disabling Spring subscription-callback lib until we get compatible Spring Boot release.


dariuszkuc:
Failure is due to the change in `graphql-java` printer that no longer includes deprecation applications IF schema does not include built-in `@deprecated` directive definition.

dariuszkuc:
NOTE: large size of the PR is due to cleaning up how we store the test schemas. Actual breaking changes from `graphql-java` updates are localized to `CacheControlInstrumentation` (as it was using old deprecated instrumentation methods) and `ServiceSDLPrinter` (as there was a change in printing behavior of `@deprecated` directive as filtering built-in directives will now result in removal all deprecation notices).

PR #403, opened 2024-05-29
fix: add missing SCHEMA `@tag` location

dariuszkuc:
Starting with Federation v2.3 `@tag` is applicable on `SCHEMA` as well.


dariuszkuc:
Updated to `minor` as while it is a fix it is also a new feature (adding new `SCHEMA` location)

PR #409, opened 2024-06-07
federation 2.8 support

clenfest:


graphql-java-support/src/main/resources/definitions_fed2_8.graphqls
```diff
@@ -0,0 +1,117 @@
...
+# federation-v2.7
+#
+
+directive @override(from: String!, label: String) on FIELD_DEFINITION
+
+#
+# federation-v2.8
+#
+
+scalar context__ContextFieldValue
```
dariuszkuc, line 113: shouldn't this be `ContextFieldValue`? (which will auto generate prefix `federation__`?)

dariuszkuc (approved):

PR #417, opened 2024-08-29
Support fed2.9

tninesling:


dariuszkuc (changes requested):
can you also add a test to verify that trying to use `@cost` with old federation version errors out?

see https://github.com/apollographql/federation-jvm/pull/377/files

graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/directives/LinkDirectiveProcessor.java
```diff
@@ -32,7 +32,9 @@ public final class LinkDirectiveProcessor {
           "@interfaceObject", 23,
           "@authenticated", 25,
           "@requiresScopes", 25,
-          "@policy", 26);
+          "@policy", 26,
+          "@cost", 29,
+          "@listSize", 29);
```
dariuszkuc, line 37: can you add missing `@context`/`@fromContext` (v2.8)?

PR #435, opened 2025-04-09
feat: add ability to propagate HTTP headers with callbacks

dariuszkuc:
Adds a mechanism for propgating HTTP headers from the original GraphQL subscription request to the subsequent callback responses.

```java
@Bean
public CallbackWebGraphQLInterceptor callbackGraphQlInterceptor(
    SubscriptionCallbackHandler callbackHandler) {
  var headers = Set.of("X-Custom-Header", "X-Custom-Header2");
  return new CallbackWebGraphQLInterceptor(callbackHandler, headers);
}
```


clenfest (approved):
The only thought I had was that was that someone might want to map the headers to a different name, but if it's not been asked for, it's probably fine to omit it.

dariuszkuc:
Yeah that would be a separate feature request

